PR #22, opened 2022-05-05
Add container image scan as part of build test pipeline

JAORMX:
This will run the [grype](https://github.com/anchore/grype)
vulnerability scanner to ensure the container images are in good shape.


JAORMX:
@joelrebel @diogomatsubara mind getting back to this? It's currently not failing on findings, but it's a good start for us to re-enable the whole pipeline later when we migrate off centos.

PR #30, opened 2022-05-11
docker: pin to Alma release 8

joelrebel:
# What does this PR do

pins Alma linux to release 8

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,10 @@
-ARG ALMA_RELEASE=8
-FROM almalinux:${ALMA_RELEASE} AS stage0
+FROM almalinux:8-minimal AS stage0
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM almalinux:8 AS stage0
+ARG ALMA_RELEASE=8
```
JAORMX, line 1: I'm actually unsure if Renovate can detect this... I would rather just be explicit and specify `almalinux:8` since it's only two occurrences.

JAORMX (changes requested):

joelrebel:
https://github.com/metal-toolbox/ironlib/runs/6389521932?check_suite_focus=true#step:6:7177
https://github.com/metal-toolbox/ironlib/runs/6387908861?check_suite_focus=true#step:6:7177

no real difference in CVE count, although a 100MB reduction in size, I'll take that for now.

PR #40, opened 2022-06-10
ci: rework image release

joelrebel:
Updates the CI image release process for development builds and image tags.

@JAORMX do you have any suggestions on how this could be tested without merging, since it's triggered off tags.

PR #41, opened 2022-06-10
Change regexp for tags

diogomatsubara:
# What does this PR do

Changes the release workflow regex tag to match the correct version

# The HW vendor this change applies to (if applicable)
n/a
# The HW model number, product name this change applies to (if applicable)
n/a
# The BMC firmware and/or BIOS versions that this change applies to (if applicable)
n/a
# What version of tooling - vendor specific or opensource does this change depend on (if applicable)
n/a
# How can this change be tested by a PR reviewer?
Should run CI and build on tagged branches

# Description for changelog/release notes
Change release workflow regex to match on correct tags

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,8 +3,8 @@ name: Release
 on:
   push:
     tags:
-      - 'v.*.*.*'
-      - 'v.*.*.*-staging'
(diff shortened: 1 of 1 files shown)


joelrebel:
Thanks, I've included this change in PR #40

PR #48, opened 2022-06-20
common.Device: replace model.Device with common.Device

joelrebel:
In the previous set of changes - 408ed07, the common.Device{} object
was embedded in model.Device{} to continue including the OemComponents field
for OEM hardware. This change moves the OemComponents field into
the Hardware{} object and so the internal model.Device{} object is removed completely
and ironlib will now return a common.Device{} for consistency with bmclib.


diogomatsubara (approved):
Thanks for this change!

PR #54, opened 2022-10-03
RAID Controller (StorageController) support; Marvell CLI (mvcli)

splaspood:
# What does this PR do

Fleshes out RAID controller support, specifically for Marvell RAID controllers.

# The HW vendor this change applies to (if applicable)

Marvell

# The HW model number, product name this change applies to (if applicable)

Marvell RAID controllers, generally.    Specifically the: 88SE9230 PCIe 2.0 x2 4-port SATA 6 Gb/s RAID Controller

# The BMC firmware and/or BIOS versions that this change applies to (if applicable)

N/A

# What version of tooling - vendor specific or opensource does this change depend on (if applicable)

mvcli (mvcli-4.1.13.31_A01 was used during development testing)

# How can this change be tested by a PR reviewer?

# Description for changelog/release notes

* StorageController extended to support RAID controller specific bits
* actions.CreateVirtualDisk and actions.DestroyVirtualDisk interfaces


joelrebel:
@splaspood if you rebase this branch on main, the linting should go through.

actions/inventory.go
```diff
@@ -50,7 +50,7 @@ func InitCollectors(trace bool) *Collectors {
 //
 // The lshw collector always executes first and is included by default.
 // nolint:gocyclo //since we're collecting inventory for each type, this is cyclomatic
-func Collect(ctx context.Context, device *common.Device, collectors *Collectors, trace, failOnError bool) error {
+func Collect(ctx context.Context, device *common.Device, collectors *Collectors, trace, failOnError bool, dynamic bool) error {
```
joelrebel, line 53: suggestion - the bool parameter type definition can be collapsed
```
func Collect(ctx context.Context, device *common.Device, collectors *Collectors, trace, failOnError, dynamic bool)
```

going ahead, I think this should use the options pattern https://github.com/tmrts/go-patterns/blob/master/idiom/functional-options.md

actions/inventory.go
```diff
@@ -119,12 +119,39 @@ func Collect(ctx context.Context, device *common.Device, collectors *Collectors,
 		return errors.Wrap(err, "error retrieving TPM inventory")
 	}
 
+	if dynamic {
+		// Update StorageControllerCollectors
+		for _, sc := range device.StorageControllers {
+			switch common.FormatVendorName(sc.Vendor) {
```
joelrebel, line 125: I suggest moving this utility detection into `utils/dynamic.go` as a method,

```go
CollectorBySlugVendor(componentSlug, vendor string) collectors.StorageControllers {
  switch componentSlug {
     case common.SlugStorageController:
          switch common.FormatVendorName(vendor) {
             case common.VendorMarvell:
          }
  }
}
```

actions/inventory.go
```diff
@@ -119,12 +119,39 @@ func Collect(ctx context.Context, device *common.Device, collectors *Collectors,
...
 	// Collect StorageController info
 	err = StorageController(ctx, device.StorageControllers, collectors.StorageControllers)
 	if err != nil && failOnError {
 		return errors.Wrap(err, "error retrieving StorageController inventory")
 	}
 
+	if dynamic {
+		for _, sc := range device.StorageControllers {
+			if sc.SupportedRAIDTypes != "" {
+				switch common.FormatVendorName(sc.Vendor) {
```
joelrebel, line 142: move this into the utils package - as mentioned in a previous comment.

actions/inventory.go
```diff
@@ -119,12 +119,39 @@ func Collect(ctx context.Context, device *common.Device, collectors *Collectors,
...
+		for _, sc := range device.StorageControllers {
+			switch common.FormatVendorName(sc.Vendor) {
+			case common.VendorMarvell:
+				collectors.StorageControllers = utils.NewMvcliCmd(trace)
+				// TODO(splaspood) We need to handle multiple collectors here
+			}
+		}
+	}
+
 	// Collect StorageController info
```
joelrebel, line 133: Could these dynamic bits be moved into the `StorageController()`, `Drive()` if those methods accepted the `dynamic` parameter?

utils/lshw.go
```diff
@@ -498,6 +508,32 @@ func (l *Lshw) xStorageController(node *LshwNode) *common.StorageController {
...
+	}
+
+	// Remove the PCIId from the sc.Model if present.
+	if matches := pciIdRegex.FindStringSubmatch(sc.Model); matches != nil {
+		sc.Model = pciIdRegex.ReplaceAllString(sc.Model, "")
+	}
+
+	// If no serial number is present, derive the serial from the PCI vendor/product id
+	if sc.Serial == "" {
+		sc.Serial = sc.Common.PCIVendorID + ":" + sc.Common.PCIProductID
```
joelrebel, line 533: 👍

utils/lshw.go
```diff
@@ -348,6 +350,14 @@ func (l *Lshw) xNIC(node *LshwNode) *common.NIC {
 		BusInfo:     node.Businfo,
 	}
 
+	// Parse out PCI Vendor and Product identifiers from node.ProductName if available
+	pciIdRegex, _ := regexp.Compile(` \[(\S{4}):(\S{4})\]$`)
+	if matches := pciIdRegex.FindStringSubmatch(nic.ProductName); matches != nil {
+		nic.ProductName = pciIdRegex.ReplaceAllString(nic.ProductName, "")
+		nic.Common.PCIVendorID = matches[1]
+		nic.Common.PCIProductID = matches[2]
```
joelrebel, line 358: A len() check here would be ideal, since this could panic if `len(matches) == 1`

utils/lshw.go
```diff
@@ -498,6 +508,32 @@ func (l *Lshw) xStorageController(node *LshwNode) *common.StorageController {
 		PhysicalID:               node.Physid,
 		BusInfo:                  node.Businfo,
 	}
+
+	// Parse out PCI Vendor and Product identifiers from node.ProductName if available
+	pciIdRegex, _ := regexp.Compile(` \[(\S+):(\S+)\]$`)
+	if matches := pciIdRegex.FindStringSubmatch(sc.ProductName); matches != nil {
+		sc.ProductName = pciIdRegex.ReplaceAllString(sc.ProductName, "")
+		sc.Common.PCIVendorID = matches[1]
+		sc.Common.PCIProductID = matches[2]
```
joelrebel, line 517: A len() check here would be ideal, since this could panic if `len(matches) == 1`

utils/mvcli.go
```diff
@@ -15,25 +15,9 @@ import (
...
-	"1":      "Mirroring (RAID-1)",
-	"10":     "Mirrored Stripe (RAID-10)",
-	"5":      "RAID-5",
-	"1e":     "RAID-1e",
-	"hc":     "hc",
-	"hs":     "hs",
-	"hybrid": "hybrid",
-}
-
+var validRaidModes = []string{"0", "1", "10", "5", "1e", "hc", "hs", "hybrid"}
```
joelrebel: The previous map that described each of RAID levels was useful to know what these identifiers are,
suggestion: leave a comment above the Modes, Types with what they mean.

Should this be called `validRaidLevels` instead ?
```
// 0: Striping (RAID-0)
// 1: Mirroring (RAID-1)
```

utils/misc.go
```diff
@@ -26,6 +26,17 @@ func StringInSlice(str string, sl []string) bool {
 	return false
 }
 
+// IntInSlice returns a bool value if the given int was not/found in the given slice.
```
joelrebel, line 29: This could be removed in favour of `slices.Contains` https://pkg.go.dev/golang.org/x/exp/slices#Contains

utils/misc.go
```diff
@@ -37,6 +37,17 @@ func IntInSlice(str int, sl []int) bool {
 	return false
 }
 
+// UintInSlice returns a bool value if the given unsigned int was not/found in the given slice.
+func UintInSlice(str uint, sl []uint) bool {
```
joelrebel, line 41: replace implementation with https://pkg.go.dev/golang.org/x/exp/slices#Contains

utils/mvcli.go
```diff
@@ -239,10 +239,8 @@ func (m *Mvcli) parseMvcliInfoHbaOutput(b []byte) []*MvcliDevice {
 	return devices
 }
 
-func stringToInt(s string) int64 {
-	const bitSize = 64
-	i, _ := strconv.ParseInt(s, 0, bitSize)
-
+func stringToInt(s string, b int) int64 {
```
joelrebel, line 242: suggestion: rename method to `stringToInt64`

joelrebel (changes requested):
Thanks for the work here, looks pretty good overall 🚀 

Could you add some examples for vdisk create, delete in https://github.com/metal-toolbox/ironlib/tree/main/examples

I've left a few change suggestions.

utils/mvcli.go
splaspood: Level.. Mode...  Type...  depends on who you ask, I suppose.   CreateVirtualDisk() and others currently expect RaidMode.

Some of these valid (by marvell standards) 'modes' are probably not 'standard' industry wide (like RAID-1, 5, 10, etc are) so I wasn't thinking we'd 'expose' those at a higher level, which is why I felt it ok to ditch the map and simplify things.  I could add comments, I suppose.

utils/misc.go
splaspood: I'd love to...   but isn't that still an 'experimental' extension to 1.18?

actions/inventory.go
splaspood: that function, as you defined it returns a var that implements collectors.StorageControllers -- should that be something that references collectors generically, since we take a componentSlug as the first param and this function could then return a collector for something other than StorageController?

joelrebel: Thats a good point, although since currently theres no generic collector type, it is simpler to have methods returning the specific type instead. I should have named that method `StorageControllerCollector instead`.

utils/mvcli.go
joelrebel: A few comments for the modes would suffice

utils/misc.go
joelrebel: True that, although, it seems like that package will make it out of experimental https://github.com/golang/go/issues/45955

Feel free to ignore my suggestion here.

splaspood: I love your suggestion here!    Would you be comfortable rolling the experimental stuff into production?  If so I'll stick with slices.Contains() (I've already made the changes in my local branch)

joelrebel: Yeah sounds good, I'm fine with having the slices experimental package as a dependency, its pretty mature at this point.

joelrebel (approved):
LGTM, thanks again for all the work here and squashing the commits!

PR #56, opened 2022-10-18
Multiple drive collectors

joelrebel:
### What does this PR do

Update actions inventory collection to support registering and executing multiple drive collectors.


psteinbachs (approved):
Looks good.

PR #57, opened 2022-10-18
Drive features

psteinbachs:
### What does this PR do

This PR adds drive feature/capability parsing for nvme and SATA drives, and an lsblk util for determining transport method of a given drive.

This change will add the following tool dependencies:
* lsblk
* hdparm

### DIFF
--- a/utils/hdparm.go
+++ b/utils/hdparm.go
@@ -44,7 +44,7 @@ func (h *Hdparm) ListFeatures(device string) ([]byte, error) {
 }
 
 // nolint:gocyclo // line parsing is cyclomatic
-func (h *Hdparm) parseHdparmFeatures(device string) ([]hdparmDeviceFeatures, error) {
+func (h *Hdparm) ParseHdparmFeatures(device string) ([]hdparmDeviceFeatures, error) {
 	out, err := h.ListFeatures(device)
 	if err != nil {
 		return nil, err
--- a/utils/hdparm.go
+++ b/utils/hdparm.go
@@ -1,0 +1,162 @@
+package utils
+
+import (
+	"bufio"
+	"context"
+	"regexp"
+	"strings"
+)
+
+const hdparm = "/usr/sbin/hdparm"
+
+type Hdparm struct {
+	Executor Executor
+}
+
+type hdparmDeviceFeatures struct {
+	Name        string `json:"Name"`
+	Description string `json:"Description"`
+	Enabled     bool   `json:"Enabled"`
+}
+
+// Return a new hdparm executor
+func NewHdparmCmd(trace bool) *Hdparm {
+	e := NewExecutor(hdparm)
+	e.SetEnv([]string{"LC_ALL=C.UTF-8"})
+
+	if !trace {
+		e.SetQuiet()
+	}
+
+	return &Hdparm{Executor: e}
+}
+
+func (h *Hdparm) ListFeatures(device string) ([]byte, error) {
+	// hdparm -I devicepath
+	h.Executor.SetArgs([]string{"-I", device})
+
+	result, err := h.Executor.ExecWithContext(context.Background())
+	if err != nil {
+		return nil, err
+	}
+
+	return result.Stdout, nil
+}
+
+func (h *Hdparm) parseHdparmFeatures(device string) ([]hdparmDeviceFeatures, error) {
+	out, err := h.ListFeatures(device)
+	if err != nil {
+		return nil, err
+	}
+
+	var features []hdparmDeviceFeatures
+
+	var lines []string
+
+	s := string(out)
+
+	scanner := bufio.NewScanner(strings.NewReader(s))
+	for scanner.Scan() {
+		lines = append(lines, scanner.Text())
+	}
+
+	err = scanner.Err()
+	if err != nil {
+		return nil, err
+	}
+
+	// Delimiters
+	feat_start := "Enabled"
+	sec_start := "Security:"
+	sec_end := "Logical Unit"
+
+	supported := regexp.MustCompile(`(?s)^supported$`)
+	seu := strings.NewReplacer("min", "")
+	sfi := strings.NewReplacer("_", " ", "-", " ", "{", "", "}", "", "(", "", ")",
+		"", ",", "", "|", "", "set", "", "command", "")
+	feat_bool, sec_bool := false, false
+
+	for _, line := range lines {
+		line := strings.TrimSpace(line)
+		parts := strings.Fields(line)
+		var flag string
+
+		// start/end match specific block delimiters
+		// bools are toggled to indicate lines within a given block
+		switch {
+		case strings.Contains(line, feat_start):
+			feat_bool = true
+		case strings.Contains(line, sec_start):
+			feat_bool, sec_bool = false, true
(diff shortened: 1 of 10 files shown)


utils/hdparm.go
```diff
@@ -44,7 +44,7 @@ func (h *Hdparm) ListFeatures(device string) ([]byte, error) {
 }
 
 // nolint:gocyclo // line parsing is cyclomatic
-func (h *Hdparm) parseHdparmFeatures(device string) ([]hdparmDeviceFeatures, error) {
+func (h *Hdparm) ParseHdparmFeatures(device string) ([]hdparmDeviceFeatures, error) {
```
joelrebel, line 47: Does this method need to be exported, should it be lower cased instead?

utils/nvme.go
```diff
@@ -131,8 +131,8 @@ func (n *Nvme) ListFeatures(device string) ([]byte, error) {
 }
 
 // nolint:gocyclo // line parsing is cyclomatic
-func (n *Nvme) parseNvmeFeatures(d *nvmeDeviceAttributes) ([]nvmeDeviceFeatures, error) {
-	out, err := n.ListFeatures(d.DevicePath)
+func (n *Nvme) ParseNvmeFeatures(device string) ([]nvmeDeviceFeatures, error) {
```
joelrebel, line 134: Does this method need to be exported, should it be lower cased instead?

utils/nvme.go
```diff
@@ -131,8 +131,8 @@ func (n *Nvme) ListFeatures(device string) ([]byte, error) {
...
+	out, err := n.ListFeatures(device)
```
joelrebel, line 135: suggestion: name the `device` parameter `devicePath` instead?

utils/lsblk.go
```diff
@@ -0,0 +1,107 @@
...
+		dModel := d.Model
+
+		var vendor string
+		var product []string
+
+		modelTokens := strings.Split(d.Model, " ")
+
+		if len(modelTokens) > 1 {
+			vendor = modelTokens[1]
+			product = append(modelTokens[:1], modelTokens[2:]...)
```
joelrebel, line 71: The usage of `append()` here is incorrect, I'm not completely sure whats going on here here - but you most likely want to call append twice.

```
product = append(product, modelTokens[:1])
product = append(product, modelTokens[2:]...) 
```

Would be worth leaving a comment as to what this block is doing.

joelrebel (commented):
Thanks for the work on these features 👍 

I've left a few comments, with those fixed it should be good to go.

utils/nvme.go
```diff
@@ -97,3 +117,119 @@ func (n *Nvme) List() ([]byte, error) {
...
+		case sani_start:
+			sani_bool = true
+		case sani_end:
+			sani_bool = false
+		}
+
+		if fna_start || sani_start {
+			var feature nvmeDeviceFeatures
+			parts := strings.Split(line, ":")
+			key, value := strings.TrimSpace(parts[0]), strings.TrimSpace(parts[1])
```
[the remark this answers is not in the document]
psteinbachs: I've added the length check here, and it's checking for exactly length with a variable because magic numbers are apparently bad.

utils/nvme.go
```diff
@@ -97,3 +117,119 @@ func (n *Nvme) List() ([]byte, error) {
...
+			if value != "0" {
+				feature.Enabled = true
+			}
+			features = append(features, feature)
+
+			// crypto erase
+		} else if fna_bool && !fna_end && !is_blank {
+			var feature nvmeDeviceFeatures
+			parts := strings.Split(line, ":")
+			data := strings.Split(parts[2], "\t")
```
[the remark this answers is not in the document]
psteinbachs: I've added the length check here, and it's checking for exactly length with a variable because magic numbers are apparently bad.

utils/nvme.go
```diff
@@ -97,3 +117,119 @@ func (n *Nvme) List() ([]byte, error) {
...
+			enabled := strings.TrimSpace(data[0])
```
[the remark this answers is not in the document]
psteinbachs: I've added the length check here, and it's checking for exactly length with a variable because magic numbers are apparently bad.

utils/nvme.go
psteinbachs: The plan was to export it in both nvme and hdparm if we were calling from lsblk and using that for the collection, but we decided against it.  I've reverted it.

utils/lsblk.go
psteinbachs: This was a poorly executed attempt to remove the first element (vendor) and use the rest as the product name.  Since lshw will be populating this, we decided lsblk doesn't need to.  This has been removed.

utils/nvme.go
```diff
@@ -97,3 +117,119 @@ func (n *Nvme) List() ([]byte, error) {
...
+	}
+
+	// Delimiters
+	re_fna_start := regexp.MustCompile(`(?s)^fna\s`)
+	re_fna_end := regexp.MustCompile(`(?s)^vwc\s`)
+	re_sani_start := regexp.MustCompile(`(?s)^sanicap\s`)
+	re_sani_end := regexp.MustCompile(`(?s)^hmminds\s`)
+	re_blank := regexp.MustCompile(`(?m)^\s*$`)
+
+	fna_bool, sani_bool := false, false
```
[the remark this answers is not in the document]
psteinbachs: Updated in both nvme and hdparm.

utils/nvme.go
```diff
@@ -97,3 +117,119 @@ func (n *Nvme) List() ([]byte, error) {
...
+				feature.Enabled = true
+			}
+			feature.Description = data[1]
+			features = append(features, feature)
+			// sanitize
+		} else if sani_bool && !sani_end && !is_blank {
+			var feature nvmeDeviceFeatures
+			var flag string
+			parts := strings.Split(line, ":")
+			data := strings.Split(parts[2], "\t")
```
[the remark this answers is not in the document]
psteinbachs: I've added the length check here, and it's checking for exact length with a variable because magic numbers are apparently bad.

utils/hdparm.go
psteinbachs: The plan was to export it in both nvme and hdparm if we were calling from lsblk and using that for the collection, but we decided against it. I've reverted it.

utils/nvme.go
psteinbachs: This has been updated everywhere.

utils/nvme.go
```diff
@@ -97,3 +117,119 @@ func (n *Nvme) List() ([]byte, error) {
...
+			// crypto erase
+		} else if fna_bool && !fna_end && !is_blank {
+			var feature nvmeDeviceFeatures
+			parts := strings.Split(line, ":")
+			data := strings.Split(parts[2], "\t")
+			enabled := strings.TrimSpace(data[0])
+
+			// Generate short flag identifier
+			for _, word := range strings.Fields(data[1]) {
+				feature.Name += strings.ToLower(word[0:1])
```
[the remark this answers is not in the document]
psteinbachs: We decided a check wasn't necessary here since the additional variables are being derived from one that is now being checked.

joelrebel (approved):
LGTM, thanks for all the work here!

PR #58, opened 2022-10-18
Logical name

joelrebel:
### What does this PR do

- Includes the logical names for drive components in the device inventory.


psteinbachs (approved):
ship it

PR #60, opened 2022-10-27
Capability collector

joelrebel:
### What does this PR do

The `actions.Collect` inventory method now returns drive capabilities.

Based on @psteinbachs work to add `hdparm`, `nvme` drive feature collection,
this change set defines and implements the `DriveCapabilities()` interface. 

- The `Device.Common` struct was extended to include `[]*Capabilities` that are then
returned by the `hdparm`, `nvme` drive capability collectors.
- Since `Capabilities` seemed a more apt name instead of `Features` the `DriveCapability` methods in the `nvme`, `hparm`  were renamed and a `[]*common.Capability` is returned.
- Context propagation for `nvme`, `hdparm` command execution.
- Lower case methods that are not exported.
- Tests updated.

This change depends on https://github.com/bmc-toolbox/common/pull/4


actions/dynamic.go
```diff
@@ -20,3 +22,26 @@ func DriveCollectorByStorageControllerVendor(vendor string, trace bool) DriveCol
...
+			return nvme
+		}
+
+		hdparm, isHdparmCollector := collector.(*utils.Hdparm)
+		if isHdparmCollector && !strings.Contains(logicalName, "nvme") {
+			return hdparm
+		}
+	}
+
+	// otherwise return the collector based on the logical name
```
psteinbachs, line 40: This is a safe assumption for our environment as discussed, but just noting it would need to be adjusted in the case of SAS or SCSI.  hdparm will at least return some basic info on SAS drives so it shouldn't fail outright.

psteinbachs (approved):
Looking good, and I agree that capabilities is more apt than features.

psteinbachs (dismissed):
This also looks good, I had meant to include it.

psteinbachs (approved):
The additional updates look solid to me, and vetUpdate() in particular is definitely easier to understand now.  Thanks for all this work!

actions/inventory.go
```diff
@@ -39,7 +40,14 @@ type Collectors struct {
 func InitCollectors(trace bool) *Collectors {
 	return &Collectors{
 		Inventory: utils.NewLshwCmd(trace),
-		Drives:    []DriveCollector{utils.NewSmartctlCmd(trace)},
+		Drives: []DriveCollector{
+			utils.NewSmartctlCmd(trace),
+			utils.NewLsblkCmd(trace),
+		},
+		DriveCapabilities: []DriveCapabilityCollector{
```
splaspood, line 47: Probably a bit late in the game for this type of question...     Why did we opt to make DriveCapabilities a 'top level' collector rather than just an additional collector (or collector(s)) added to Drives based upon the type of Drives found during the initial Inventory?

splaspood (approved):
Nothing jumped out at me as being obviously wrong or bad form.   I noticed us hard coding some specific capability names in utils/nvme.go and wonder if maybe that should be moved into something common, but I didn't feel it was something that would hold up approval.

actions/inventory.go
joelrebel: They are separate since not all collectors can return drives and drive capabilities,
and and so extending the existing `DriveCollectors` interface with a `DriveCapabilities()` method wouldn't be nice to maintain - it forces all DriveCollector utils to implement the `DriveCapabilities()` method even if they don't support it. 

A smaller scoped interface seems ideal for this case. Hope that answers the question?

splaspood: Yeah I think that clarifies things for me.

PR #62, opened 2022-11-03
Drive capabilities fixes

joelrebel:
### What does this PR do

Improves drive capabilities collection and drive data patching

- Fixes for lsblk Drives collector.
- Fallback to match drives by logical name if theres no serial number.
- Adds the drive capabilities collection for SMC hardware.
- Trim space from strings before comparison for the Description, ProductName.
- Make sure the Serial attribute is not overwritten.


psteinbachs (approved):
Looks good to me.

PR #69, opened 2023-01-31
Improve logging control for StorageController actions

splaspood:
### What does this PR do
Allows someone to specify the logger and thus the logging level for StorageController related actions.

Currently only Marvell storage controllers are supported by ironlib, so... only Marvell HW at the moment.


actions/storage_controller.go
```diff
@@ -9,32 +9,37 @@ import (
...
 var (
 	ErrVirtualDiskManagerUtilNotIdentified = errors.New("virtual disk management utility not identifed")
 )
 
-func CreateVirtualDisk(ctx context.Context, hba *common.StorageController, options *model.CreateVirtualDiskOptions) error {
-	util, err := GetControllerUtility(hba.Vendor, hba.Model)
+type StorageControllerAction struct {
+	Logger logrus.Logger
+}
+
```
joelrebel, line 22: Add a helper method to instantiate this object.

```go
func NewStorageControllerAction(logger *logrus.Logger) *StorageControllerAction {
  return &StorageControllerAction{logger}
}
```

joelrebel (changes requested):

joelrebel (approved):
Thanks for the fixes, LGTM!

PR #70, opened 2023-02-06
Handle new drives found by additional inventory collector(s).

splaspood:
### What does this PR do
* Handle new drives found by additional inventory collector(s)
* Ensure physical drive types are being parsed properly
* Pass entire 'device' object into Drives() collector func

Any future Drive collectors we add, specifically Marvell storage controllers in this particular case.


actions/inventory.go
```diff
@@ -187,7 +190,7 @@ func Collect(ctx context.Context, device *common.Device, collectors *Collectors,
 
 // Drives executes drive collectors and merges the data into device.[]*Drive
 // nolint:gocyclo // TODO(joel) if theres more conditionals to be added in here, the method is to be split up.
-func Drives(ctx context.Context, drives []*common.Drive, collectors []DriveCollector) error {
+func Drives(ctx context.Context, device *common.Device, collectors []DriveCollector) error {
```
joelrebel, line 193: This is cleaner than what I had in mind - passing a pointer to the drive slice, I had to check what the code would look like, and it wasn't pretty https://go.dev/play/p/lw5RB8lXdfR

utils/mvcli.go
```diff
@@ -338,6 +338,8 @@ func (m *Mvcli) processDriveType(pdType, ssdType string) string {
 		return common.SlugDriveTypeSATASSD
 	}
 
+	fmt.Println("Got to UNKNOWN")
```
joelrebel, line 341: purge debug print ? :)

splaspood: Debug prints are the seasoning of true champions ;)   Yeah will do.

joelrebel (approved):
🚀

PR #71, opened 2023-02-17
Allow collectors to be disabled at runtime via environment var

splaspood:
### What does this PR do

Allow collectors to be disabled at runtime via IRONLIB_DISABLED_COLLECTORS environment var

* Add IRONLIB_DISABLED_COLLECTORS env var
* Add isCollectorDisabled(collector string) bool
* Add isThisCollectorDisabled() bool that uses runtime to figure out the calling function name (which would be the "collector name"
* Add call to isThisCollectorDisabled() to all top level collectors Drives, BMC, etc.

Set IRONLIB_DISABLED_COLLECTORS to a comma separated list of collector names. 

eg.  IRONLIB_DISABLED_COLLECTORS=DriveCapabilities,StorageController

Allow collectors to be disabled at runtime via IRONLIB_DISABLED_COLLECTORS environment var

### DIFF
--- a/actions/inventory.go
+++ b/actions/inventory.go
@@ -3,6 +3,8 @@ package actions
 import (
 	"context"
 	"log"
+	"os"
+	"runtime"
 	"strings"
 
 	"github.com/bmc-toolbox/common"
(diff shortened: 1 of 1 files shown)


splaspood:
I got kinda 'clever' with the isThisCollectorDisabled() function... I don't know if you (@joelrebel) have any negative opinions on such things..

If not I'm wondering if we shouldn't make a validateCollector() function that calls that and also handles the if collectorSlice == nil return bit so we can stop repeating that in each primary collector func.   I'll need to bring down the complexity of the StorageController func.

splaspood:
And as of this comment I haven't tested this code at all.

joelrebel:
Since theres a need to pass more and more options to the library, we need a better way than env vars and lots of function parameters, I'm reworking the way options are passed into the library, should be done in the next day or so, hold on a bit until this is through https://github.com/metal-toolbox/ironlib/tree/options

splaspood:
Closing this for the Options PR @joelrebel proposed.

PR #75, opened 2023-02-21
Options

joelrebel:
### What does this PR do

- `GetInventory()` now accepts an arbitrary number of parameters that were earlier set in different parts of the codebase.
- Refactors the Actions methods to be under a type so parameters can be shared among the collection methods.
- Removes duplicate interfaces.
- Updates providers to conform to updated interfaces.

- Run the inventory example.

- Actions refactor


actions/inventory.go
```diff
@@ -1,54 +1,136 @@
...
+	failOnError bool
+
+	// dynamicCollection when enabled will cause ironlib to
+	// enable collectors based on the detected component vendor.
+	dynamicCollection bool
+
+	// disableCollectorUtilities is the list of collector utilities
+	// to be disabled, this is the name of collector utility
+	// which is returned by its Attributes() method.
+	disableCollectorUtilities []model.CollectorUtility
```
splaspood, line 45: Maybe this should be 'disabledCollectorUtilities' ?

actions/inventory.go
```diff
@@ -1,54 +1,136 @@
...
+
+// WithCollectors sets collectors to the ones passed in as a parameter.
+func WithCollectors(collectors *Collectors) Option {
+	return func(a *InventoryCollectorAction) {
+		a.collectors = *collectors
+	}
+}
+
+// WithDisableCollectorUtilities disables the given collector utilities.
+func WithDisableCollectorUtilities(utilityNames []model.CollectorUtility) Option {
```
splaspood, line 98: s/WithDisableCollectorUtilities/WithDisabledCollectorUtilities/g

actions/inventory.go
```diff
@@ -353,178 +464,218 @@ func NICs(ctx context.Context, nics []*common.NIC, c NICCollector) error {
...
 	if err != nil {
 		return err
 	}
 
-	if len(ncontrollers) == 0 {
+	if len(found) == 0 {
 		return nil
 	}
 
 	// TODO: handle case where the object may not already be present in device.Controllers and needs to be added
```
splaspood, line 668: Is this still the case?  If so, do we want to pull in the code added to the Drive collector bit into here?  Maybe into a function that they all can call?

splaspood (changes requested):

actions/inventory.go
splaspood: This was supposed to be on the TODO comment, not that if... not sure how I fouled that up.

actions/inventory.go
joelrebel: sounds right, fixed in[2057e09](https://github.com/metal-toolbox/ironlib/pull/75/commits/2057e0930e7282cb218d0230f39567138b082fd3)

actions/inventory.go
joelrebel: fixed in [2057e09](https://github.com/metal-toolbox/ironlib/pull/75/commits/2057e0930e7282cb218d0230f39567138b082fd3)

actions/inventory.go
joelrebel: @splaspood yep that did cross my mind, I can PR that change after this gets merged though, since the scope of the changes here is quite large already.

splaspood (approved):

PR #76, opened 2023-02-23
Options and multiple storagecontrollers

joelrebel:
### What does this PR do

- Fixes trace level logging option not being set on collector utilities
- Implements support for multiple storage controller collectors

- Support multiple storage controller collectors


splaspood (commented):
Looks good.   Do we want to do similar things for NICs (and possibly other types)?

joelrebel:
> Looks good. Do we want to do similar things for NICs (and possibly other types)?

yep! although lets do that on a case by case basis

splaspood (approved):

PR #77, opened 2023-02-27
start dell helper service that loads ipmi modules

diogomatsubara:
### What does this PR do
start the dell service that loads ipmi modules

idracadm7 relies on a properly configured
/opt/dell/srvadmin/etc/srvadmin-hapi/ini/dchipm.ini file, starting the helper before collecting bios configuration, makes sure the dchipm.ini file is populated and ipmi modules are loaded.

Dell

The easiest way is to run an alloy-docker image (which is based on the ironlib docker image) on a dell system


joelrebel (approved):
Thanks!

PR #79, opened 2023-02-28
StorageController and Drives collector fixes

splaspood:
### What does this PR do

Corrects the handling of "dynamic" StorageController and Drive collectors. 

All StorageController and Drive dynamic collectors which at this point is just Marvell (mvcli).

### DIFF
--- a/actions/inventory.go
+++ b/actions/inventory.go
@@ -796,7 +796,7 @@ func (a *InventoryCollectorAction) vetUpdate(change *diff.Change) bool {
 			return true
 		}
 	case int:
-		if newValue != int(0) {
+		if newValue >= -1 {
 			return true
 		}
 	case int64:
--- a/actions/inventory.go
+++ b/actions/inventory.go
@@ -250,7 +250,9 @@ func (a *InventoryCollectorAction) Collect(ctx context.Context, device *common.D
 		for _, sc := range a.device.StorageControllers {
 			c := DriveCollectorByStorageControllerVendor(sc.Vendor, a.trace)
 
-			if c != nil {
+			if c := DriveCollectorByStorageControllerVendor(sc.Vendor, a.trace); c != nil {
+                a.collectors.DriveCollectors = append(a.collectors.DriveCollectors, c)
+            }
 				a.collectors.DriveCollectors = append(a.collectors.DriveCollectors, c)
 			}
 		}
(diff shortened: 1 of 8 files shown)


splaspood:
Still working through the test problems, not yet clear why non Marvell attached drives are winding up with a StorageControllerDriveID of 0 (should be -1).

Possibly controversial changes in this PR:

* We normalize the StorageController Vendor by passing it through common.FormatVendorName().    I think if this PR is accepted we should do this for all hardware...

splaspood:
I believe I have fixed the tests, but can't figure out why github actions is complaining during linting.. can't repro elsewhere.

actions/inventory.go
```diff
@@ -229,10 +229,11 @@ func (a *InventoryCollectorAction) Collect(ctx context.Context, device *common.D
 	// Update StorageControllerCollectors based on controller vendor attributes
 	if a.dynamicCollection {
 		for _, sc := range a.device.StorageControllers {
-			a.collectors.StorageControllerCollectors = append(
-				a.collectors.StorageControllerCollectors,
-				StorageControllerCollectorByVendor(sc.Vendor, a.trace),
-			)
+			c := StorageControllerCollectorByVendor(sc.Vendor, a.trace)
```
joelrebel, line 232: This would be a bit easier to read in this form,

```suggestion
		if c := StorageControllerCollectorByVendor(sc.Vendor, a.trace); c != nil {
            a.collectors.StorageControllerCollectors = append(a.collectors.StorageControllerCollectors, c)
        }
```

actions/inventory.go
```diff
@@ -245,15 +246,14 @@ func (a *InventoryCollectorAction) Collect(ctx context.Context, device *common.D
...
-			a.collectors.DriveCollectors = append(
-				a.collectors.DriveCollectors,
-				DriveCollectorByStorageControllerVendor(sc.Vendor, a.trace),
-			)
+			if c != nil {
+				a.collectors.DriveCollectors = append(a.collectors.DriveCollectors, c)
+			}
+		}
 
+		if len(a.collectors.DriveCollectors) > 0 {
```
joelrebel, line 256: return early so we avoid larger nested blocks of conditionals

```suggestion
		if len(a.collectors.DriveCollectors) == 0 {
            continue
        }
```

joelrebel (changes requested):
Thanks for these fixes, left a few minor suggestions. Otherwise looks good to me!

actions/inventory.go
splaspood: Since this code isn't within a for loop, and there's additional stuff further on in the function I wouldn't want to skip, I think this code is best as-is.

joelrebel (approved):

PR #84, opened 2023-04-04
Add Dockerfile and dependencies script

diogomatsubara:
Dockerfile will build an ironlib image with only the opensource dependencies and provide a way for downstream users to include extra dependencies for private images.


Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+    hdparm        \
+    util-linux    \
+    nvme-cli      \
+    dmidecode     \
+    libssh2-devel \
+    kmod          \
+    tar           \
+    gzip          \
+    unzip         \
+    which         \
```
mmlb, line 57: I find it very nice to indent the long-list-of-things for a command by a couple of spaces so that the indentation helps with grouping.
So something like:
```
RUN microdnf ... \
      lshw \
      ...
      smartmontools && \
    gdisk && \
    ...
```
Also sorting long lists is nice too :D

dependencies.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
mmlb, line 1: when is this file used?

Dockerfile
mmlb: Oh another thing that I've found to be nice when having a long list in the middle of && \ chains in Dockerfiles is to end the list with a solo `;` this way you can add new things at the end and its only one line of diff:

```
RUN microdnf ... \
      lshw \
      ...
      smartmontools \
      ; && \
    gdisk && \
```

Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+    smartmontools && \
+    gdisk && \
+    e2fsprogs && \
+    dosfstools && \
+    mdadm && \
```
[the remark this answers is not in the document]
diogomatsubara: They are to be installed by microdnf

Dockerfile
diogomatsubara: I'll get that fixed before the final PR is ready. I got the list broken when adding the latest changes done in original Dockerfile

dependencies.sh
diogomatsubara: It's supposed to be run before running `docker build`. I'll include more details in the README before the final PR is ready

Dockerfile
diogomatsubara: Fixed in [a3b0d66](https://github.com/metal-toolbox/ironlib/pull/84/commits/a3b0d668acd8efe2ed5c1d465685196963d0b747)

Dockerfile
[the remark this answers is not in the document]
diogomatsubara: Fixed in [a3b0d66](https://github.com/metal-toolbox/ironlib/pull/84/commits/a3b0d668acd8efe2ed5c1d465685196963d0b747)

Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+    util-linux    \
+    which         \
+    microdnf clean all && \
+    ln -s /usr/bin/microdnf /usr/bin/yum # since dell dsu expects yum
+
+# Provide hook to include extra dependencies in the image
+RUN if [[ -f ${DEPDIR}/install-extra-deps.sh ]]; then cd ${DEPDIR} && bash install-extra-deps.sh; fi
+
+# Delete /tmp/* and dependencies dir as we don't need those included in the image.
+# for some reason this is not working in reducing the size of the image
```
mmlb, line 70: The image size isn't being reduced because all this is happening across multiple RUNs and they are being stored in each layer. The best solution to that problem w/o resorting to one huge RUN is to change `FROM almalinux:9-minimal` as `FROM almalinux:9-minimal AS stage1` and keep the rest of this file as is and then later do

```
FROM scratch
COPY --from=stage1 / /
```
This will give you only the files that have been kept at the end of the last command of the last stage. Effectively squashing the image layers down to just 1.

diogomatsubara: I thought each `RUN` would create a new layer and then the last instruction would be the layer used. Apparently it doesn't work like that. Anyway, thanks for the suggestion, fixed in [dbfa026](https://github.com/metal-toolbox/ironlib/pull/84/commits/dbfa0268b18ee71b258940973c479e09eda304bf) and image is now much smaller.

dependencies.sh
diogomatsubara: I got rid of the `dependencies.sh` in [6261f8d](https://github.com/metal-toolbox/ironlib/pull/84/commits/6261f8d0a6774aed0794150e372eb040997ce162) it was only used to fetch and install the epel RPM package so I moved that to Dockerfile.

joelrebel (approved):
This looks good to me, thanks for the efforts here!

Dockerfile
```diff
@@ -0,0 +1,79 @@
...
+
+# Build a lean image with dependencies installed.
+FROM scratch
+COPY --from=stage1 / /
+
+# Provide hook to include extra dependencies in the image
+ONBUILD ARG DEPDIR="dependencies"
+ONBUILD COPY "${DEPDIR}" dependencies
+ONBUILD RUN if [[ -f ${DEPDIR}/install-extra-deps.sh ]]; then cd ${DEPDIR} && bash install-extra-deps.sh; fi
+ONBUILD RUN rm -rf "${DEPDIR}"
```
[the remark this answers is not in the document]
mmlb: Glad to see the ONBUILD works here and I think it ends up being cleaner than the first version.

PR #96, opened 2023-05-02
Add chainguard config

diogomatsubara:
### What does this PR do

Add chainguard config so renovate PRs will show up in Enforce as signed


joelrebel (approved):
nice!

PR #102, opened 2023-06-12
.github/workflows/release-docker-image.yml: Use --recursive when calling cosign sign

splaspood:
…ing cosign sign

### What does this PR do

Currently we sign our multiarch docker image without the --recursive option.   It's believed this 'breaks' our multiarch image.


diogomatsubara:
LGTM, lint failed with a timeout so I kicked another run of it as it seems to be due to some flakiness.

diogomatsubara (approved):

PR #103, opened 2023-06-12
Repush images after signing is complete

mmlb:
### What does this PR do

Re-pushes to GHCR after images are signed. Hopefully this causes GHCR to end up with a valid image manifest, up until now all of them have been weird/broken.


mmlb:
Not sure if this will actually fix the issue, but worth a try.

splaspood (approved):

PR #120, opened 2023-12-21
Vc/logo uefi image hash

DoctorVin:
Implement logo checksum on inventory


providers/supermicro/supermicro.go
```diff
@@ -98,6 +98,7 @@ func (s *supermicro) GetInventory(ctx context.Context, options ...actions.Option
 			firmware.MakeOutputPath(),
 			firmware.TraceExecution(trace),
 		),
+		UEFIVarsCollector: &utils.UEFIVariableCollector{},
```
nshalman, line 101: Is there a reason we are only doing UEFIVars collection on SMC? Could it be added to other providers too or added somewhere more generic? NOT A BLOCKER FOR LANDING!

joelrebel: its by default collected on the rest

PR #121, opened 2024-01-04
Use metal-toolbox version of flashrom

nshalman:
### What does this PR do

Use https://github.com/metal-toolbox/flashrom/releases/tag/v1.3.99

Test if the newer flashrom binary works better.

Updated flashrom with patches from metal-toolbox

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -73,9 +73,9 @@ RUN microdnf install -y --setopt=tsflags=nodocs --setopt=install_weak_deps=0 \
     pciutils      \
     smartmontools \
     tar           \
+    udev          \
     unzip         \
     util-linux    \
-    flashrom      \
     python        \
     python-devel  \
     python-pip  \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -86,6 +98,9 @@ RUN microdnf install -y --setopt=tsflags=nodocs --setopt=install_weak_deps=0 \
 
 RUN pip install uefi_firmware==v1.11
 
+# Install our custom flashrom package
+COPY --from=flashrom /root/rpmbuild/RPMS/x86_64/flashrom*.rpm /tmp
+RUN rpm -ivh /tmp/flashrom*.rpm
```
stephen-fox, line 103: I recall back in CentOS that using `yum install /path/to/rpm` was preferable to using `rpm -i`... it probably does not matter here, but it is something to think about.

stephen-fox (approved):

Dockerfile
nshalman: I was unable to figure out how to get microdnf to install from a local file. Suggestions welcome.

DoctorVin (approved):

PR #128, opened 2024-04-10
House/spring cleaning

mmlb:
### What does this PR do

Tweaks the golangci-lint config so we opt out of linters instead of in. This way when new linters are introduced we can take a moment and decided if we want to pay attention to it or disable it.

Also updates to the oldest non-EOL version of go.


mmlb:
This is a big diff, yes but can be a reviewed a little easier by looking at each commit on its own.

.github/workflows/push-pr-lint.yaml
```diff
@@ -8,7 +8,7 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v4
       with:
-        go-version: '^1.19'
+        go-version: '^1.21'
```
turegano-equinix, line 11: what's wrong with 1.22?

turegano-equinix: I like the 1.22 for loops decision https://tip.golang.org/doc/go1.22 they say that it maintains the Go 1 [promise of compatibility](https://tip.golang.org/doc/go1compat)

mmlb: Yeah and it would be useful for at least one of the things I fixed in this PR, but I’m in no hurry tbh.

joelrebel:
@mmlb thanks for the clean up here, 

We're trying to keep the linter config consistent across the rest of the metal-toolbox projects, would prefer if we went with the configuration currently in [mctl](https://github.com/metal-toolbox/mctl/blob/main/.golangci.yml) which was updated for 1.22. We're can discuss any additional linters you'd want to add in.

mmlb:
> @mmlb thanks for the clean up here,
> 
> We should be fine to update this library to 1.22 (for other metal-toolbox projects).
> 
> We're trying to keep the linter config consistent across the rest of the metal-toolbox projects, would prefer if we went with the configuration currently in [mctl](https://github.com/metal-toolbox/mctl/blob/main/.golangci.yml?rgh-link-date=2024-04-11T14%3A52%3A51Z) which was updated for 1.22.

Would it make sense to create a repo with just the golangci-lint config file and have that be the source of truth? Its not currently obvious that mctl's config file is the blessed version and if my own experiences play out again the next new project/repo will end up becoming the new source of truth :D.

> We're can discuss any additional linters you'd want to add in.

I'd like to make this config be the new source-of-truth and would be happy to make sure it works in mctl too. My favorite part of this config is that we explicitly opt out of linters. This is nice because when we do update to a newer golangci-lint that has new linters we'd be in the right time to evaluate adopting or not. This also has the benefit of actually using the linters we think we are using. The whole reason I messed around with the config here is because `gofumpt` wasn't actually enabled and when I ran it as part of my editor's auto-save hook the changes were unexpected!

mmlb:
fyi applying this .golangci.yml to mctl gives the following:

<details>
<summary>40 files changed, 175 insertions(+), 255 deletions(-)</summary>
```
git diff --stat --cached *
 cmd/collect/collect.go      |  5 ++---
 cmd/collect/inventory.go    | 13 ++++---------
 cmd/collect/status.go       |  8 +++-----
 cmd/common.go               | 33 ++++++++++++++++++---------------
 cmd/create/bomupload.go     |  9 +++------
 cmd/create/create.go        |  3 +--
 cmd/create/firmware.go      | 12 ++++--------
 cmd/create/firmware_set.go  | 19 ++++++++-----------
 cmd/create/server.go        |  9 +++------
 cmd/delete/delete.go        |  3 +--
 cmd/delete/firmware.go      | 10 ++++------
 cmd/delete/firmware_set.go  |  9 +++------
 cmd/delete/serverdelete.go  | 13 +++++--------
 cmd/edit/edit.go            |  3 +--
 cmd/edit/firmware_set.go    |  9 +++------
 cmd/flags.go                |  6 ++----
 cmd/generate/docs.go        |  4 ++--
 cmd/get/bios-config.go      | 13 +++++--------
 cmd/get/bom.go              |  9 +++------
 cmd/get/condition.go        |  9 +++------
 cmd/get/firmware-set.go     | 13 +++++--------
 cmd/get/firmware.go         | 11 ++++-------
 cmd/get/get.go              |  9 +++------
 cmd/get/server.go           | 13 ++++++++-----
 cmd/install/firmware_set.go | 20 +++++++-------------
 cmd/install/install.go      |  5 ++---
 cmd/install/status.go       |  7 +++----
 cmd/list/component.go       | 11 ++++-------
 cmd/list/firmware.go        | 17 +++++++----------
 cmd/list/firmware_set.go    | 13 +++++--------
 cmd/list/list.go            |  9 +++------
 cmd/list/server.go          |  9 ++++-----
 cmd/power/power.go          | 18 ++++++++----------
 cmd/root.go                 |  4 ++--
 cmd/types.go                |  2 +-
 cmd/version.go              |  5 ++---
 internal/app/app.go         | 17 +++++++++--------
 internal/app/clients.go     |  6 +++---
 internal/auth/oauth.go      | 39 ++++++++++++++++-----------------------
 pkg/model/model.go          |  3 +--
 40 files changed, 175 insertions(+), 255 deletions(-)
```
</details>

<details>
<summary>git diff --cached *</summary>

gochecknoinits feels like its not worth it or we might want to exlude cmd/* paths

```diff
diff --git a/cmd/collect/collect.go b/cmd/collect/collect.go
index 3091a4e..00aafd8 100644
--- a/cmd/collect/collect.go
+++ b/cmd/collect/collect.go
@@ -1,9 +1,8 @@
 package collect
 
 import (
-	"github.com/spf13/cobra"
-
 	"github.com/metal-toolbox/mctl/cmd"
+	"github.com/spf13/cobra"
 )
 
 var collect = &cobra.Command{
@@ -14,7 +13,7 @@ var collect = &cobra.Command{
 	},
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	cmd.RootCmd.AddCommand(collect)
 	collect.AddCommand(collectInventoryCmd)
 	collect.AddCommand(inventoryStatus)
diff --git a/cmd/collect/inventory.go b/cmd/collect/inventory.go
index 634e648..1f79021 100644
--- a/cmd/collect/inventory.go
+++ b/cmd/collect/inventory.go
@@ -6,13 +6,11 @@ import (
 	"log"
 
 	"github.com/google/uuid"
-	"github.com/spf13/cobra"
-
 	coapiv1 "github.com/metal-toolbox/conditionorc/pkg/api/v1/types"
-	rctypes "github.com/metal-toolbox/rivets/condition"
-
 	mctl "github.com/metal-toolbox/mctl/cmd"
 	"github.com/metal-toolbox/mctl/internal/app"
+	rctypes "github.com/metal-toolbox/rivets/condition"
+	"github.com/spf13/cobra"
 )
 
 type collectInventoryFlags struct {
@@ -21,16 +19,13 @@ type collectInventoryFlags struct {
 	skipBiosConfigCollect     bool
 }
 
-var (
-	flagsDefinedCollectInventory *collectInventoryFlags
-)
+var flagsDefinedCollectInventory *collectInventoryFlags
 
 var collectInventoryCmd = &cobra.Command{
 	Use:   "inventory",
 	Short: "Collect current server firmware status and bios configuration",
 	Run: func(cmd *cobra.Command, _ []string) {
 		collectInventory(cmd.Context())
-
 	},
 }
 
@@ -74,7 +69,7 @@ func collectInventory(ctx context.Context) {
 	log.Printf("status=%d msg=%s conditionID=%s", response.StatusCode, response.Message, condition.ID)
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	flagsDefinedCollectInventory = &collectInventoryFlags{}
 
 	mctl.AddServerFlag(collectInventoryCmd, &flagsDefinedCollectInventory.serverID)
diff --git a/cmd/collect/status.go b/cmd/collect/status.go
index 887f904..762951d 100644
--- a/cmd/collect/status.go
+++ b/cmd/collect/status.go
@@ -6,12 +6,10 @@ import (
 	"log"
 
 	"github.com/google/uuid"
-	"github.com/spf13/cobra"
-
-	rctypes "github.com/metal-toolbox/rivets/condition"
-
 	mctl "github.com/metal-toolbox/mctl/cmd"
 	"github.com/metal-toolbox/mctl/internal/app"
+	rctypes "github.com/metal-toolbox/rivets/condition"
+	"github.com/spf13/cobra"
 )
 
 type inventoryStatusParams struct {
@@ -54,7 +52,7 @@ func statusCheck(ctx context.Context) {
 	fmt.Println(s)
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	inventoryStatusFlags = &inventoryStatusParams{}
 
 	mctl.AddServerFlag(inventoryStatus, &inventoryStatusFlags.serverID)
diff --git a/cmd/common.go b/cmd/common.go
index 1d9588d..5372c64 100644
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -8,17 +8,16 @@ import (
 	"strings"
 	"time"
 
+	bmclibcomm "github.com/bmc-toolbox/common"
 	"github.com/davecgh/go-spew/spew"
 	"github.com/google/uuid"
-	"github.com/metal-toolbox/mctl/internal/app"
-	"github.com/pkg/errors"
-	"golang.org/x/net/context"
-
-	bmclibcomm "github.com/bmc-toolbox/common"
 	coapiv1 "github.com/metal-toolbox/conditionorc/pkg/api/v1/types"
 	fleetdbapi "github.com/metal-toolbox/fleetdb/pkg/api/v1"
+	"github.com/metal-toolbox/mctl/internal/app"
 	rctypes "github.com/metal-toolbox/rivets/condition"
 	rt "github.com/metal-toolbox/rivets/types"
+	"github.com/pkg/errors"
+	"golang.org/x/net/context"
 )
 
 var (
@@ -30,7 +29,7 @@ var (
 const (
 	CmdTimeout = 20 * time.Second
 
-	// TODO: merge constants along with the ones in Alloy into a separate library
+	// TODO: merge constants along with the ones in Alloy into a separate library.
 	ServerVendorAttributeNS = "sh.hollow.alloy.server_vendor_attributes"
 	FirmwareSetAttributeNS  = "sh.hollow.firmware_set.labels"
 )
@@ -58,7 +57,7 @@ func AttributeFromLabels(ns string, labels map[string]string) (*fleetdbapi.Attri
 
 // AttributeByNamespace returns the serverservice attribute in the slice that matches the namespace
 //
-// TODO: move into common library and share with Alloy
+// TODO: move into common library and share with Alloy.
 func AttributeByNamespace(ns string, attributes []fleetdbapi.Attributes) *fleetdbapi.Attributes {
 	for _, attribute := range attributes {
 		if attribute.Namespace == ns {
@@ -71,7 +70,7 @@ func AttributeByNamespace(ns string, attributes []fleetdbapi.Attributes) *fleetd
 
 // VendorModelFromAttrs unpacks the attributes payload to return the vendor, model attributes for a server
 //
-// TODO: move into common library and share with Alloy
+// TODO: move into common library and share with Alloy.
 func VendorModelFromAttrs(attrs []fleetdbapi.Attributes) (vendor, model string) {
 	attr := AttributeByNamespace(ServerVendorAttributeNS, attrs)
 	if attr == nil {
@@ -88,7 +87,7 @@ func VendorModelFromAttrs(attrs []fleetdbapi.Attributes) (vendor, model string)
 
 // FirmwareSetIDByVendorModel returns the firmware set ID matched by the vendor, model attributes
 //
-// TODO: move into common library
+// TODO: move into common library.
 func FirmwareSetIDByVendorModel(ctx context.Context, vendor, model string, client *fleetdbapi.Client) (uuid.UUID, error) {
 	fwSet, err := FirmwareSetByVendorModel(ctx, vendor, model, client)
 	if err != nil {
@@ -107,7 +106,7 @@ func FirmwareSetIDByVendorModel(ctx context.Context, vendor, model string, clien
 
 // FirmwareSetByVendorModel returns the firmware set matched by the vendor, model attributes
 //
-// TODO: move into common library
+// TODO: move into common library.
 func FirmwareSetByVendorModel(ctx context.Context, vendor, model string, client *fleetdbapi.Client) ([]fleetdbapi.ComponentFirmwareSet, error) {
 	vendor = strings.TrimSpace(vendor)
 	if vendor == "" {
@@ -171,7 +170,7 @@ func FirmwareSetByVendorModel(ctx context.Context, vendor, model string, client
 	return fwSet, nil
 }
 
-type ErrUnexpectedResponse struct {
+type ErrUnexpectedResponse struct { //nolint:errname // fine for now
 	statusCode int
 	message    string
 }
@@ -190,7 +189,7 @@ func newErrUnexpectedResponse(statusCode int, message string) error {
 	return &ErrUnexpectedResponse{statusCode, message}
 }
 
-// ConditionFromResponse returns a Condition object from the Condition API ServerResponse object
+// ConditionFromResponse returns a Condition object from the Condition API ServerResponse object.
 func ConditionFromResponse(response *coapiv1.ServerResponse) (rctypes.Condition, error) {
 	if response.StatusCode != http.StatusOK {
 		return rctypes.Condition{}, newErrUnexpectedResponse(response.StatusCode, response.Message)
@@ -222,15 +221,18 @@ func FormatConditionResponse(response *coapiv1.ServerResponse, kind rctypes.Kind
 
 	if response.Records == nil {
 		err := errors.New("no records returned")
+
 		return "", err
 	}
 
 	if len(response.Records.Conditions) == 0 {
 		err := errors.New("no record found for Condition")
+
 		return "", err
 	}
 
 	var inc *rctypes.Condition
+
 	for _, c := range response.Records.Conditions {
 		if c.Kind == kind {
 			inc = c
@@ -239,6 +241,7 @@ func FormatConditionResponse(response *coapiv1.ServerResponse, kind rctypes.Kind
 
 	if inc == nil {
 		err := errors.New("response contains no condition of type: " + string(kind))
+
 		return "", err
 	}
 
@@ -265,7 +268,7 @@ func FormatConditionResponse(response *coapiv1.ServerResponse, kind rctypes.Kind
 func PrintResults(format string, data ...any) {
 	switch format {
 	case "text":
-		spew.Dump(data)
+		spew.Dump(data...)
 	case "json", "JSON":
 		b, err := json.MarshalIndent(data, "", "  ")
 		if err != nil {
@@ -276,11 +279,11 @@ func PrintResults(format string, data ...any) {
 	}
 }
 
-// Query server BMC credentials and update the given server object
+// Query server BMC credentials and update the given server object.
 func ServerBMCCredentials(ctx context.Context, client *fleetdbapi.Client, server *rt.Server) error {
 	cred, _, err := client.GetCredential(ctx, uuid.MustParse(server.ID), fleetdbapi.ServerCredentialTypeBMC)
 	if err != nil {
-		// nolint:goerr113 // error is readable when formatted
+		//nolint:goerr113 // error is readable when formatted
 		return fmt.Errorf("error in credential lookup for: %s, err: %s", server.ID, err.Error())
 	}
 
diff --git a/cmd/create/bomupload.go b/cmd/create/bomupload.go
index 0e6197c..ee6d761 100644
--- a/cmd/create/bomupload.go
+++ b/cmd/create/bomupload.go
@@ -4,10 +4,9 @@ import (
 	"log"
 	"os"
 
-	"github.com/spf13/cobra"
-
 	mctl "github.com/metal-toolbox/mctl/cmd"
 	"github.com/metal-toolbox/mctl/internal/app"
+	"github.com/spf13/cobra"
 )
 
 // Create Bom informations.
@@ -16,9 +15,7 @@ type uploadBomFileFlags struct {
 	bomXlsxFile string
 }
 
-var (
-	flagsUploadBomFileFlags *uploadBomFileFlags
-)
+var flagsUploadBomFileFlags *uploadBomFileFlags
 
 var uploadBomFile = &cobra.Command{
 	Use:   "bom",
@@ -45,7 +42,7 @@ var uploadBomFile = &cobra.Command{
 	},
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	flagsUploadBomFileFlags = &uploadBomFileFlags{}
 	usage := "xlsx file with BOM information"
 
diff --git a/cmd/create/create.go b/cmd/create/create.go
index 783fdda..e9d0bf3 100644
--- a/cmd/create/create.go
+++ b/cmd/create/create.go
@@ -2,7 +2,6 @@ package create
 
 import (
 	"github.com/metal-toolbox/mctl/cmd"
-
 	"github.com/spf13/cobra"
 )
 
@@ -14,7 +13,7 @@ var create = &cobra.Command{
 	},
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	cmd.RootCmd.AddCommand(create)
 	create.AddCommand(createFirmware)
 	create.AddCommand(createFirmwareSet)
diff --git a/cmd/create/firmware.go b/cmd/create/firmware.go
index 0a737ba..691cf7b 100644
--- a/cmd/create/firmware.go
+++ b/cmd/create/firmware.go
@@ -6,22 +6,18 @@ import (
 	"os"
 
 	fleetdbapi "github.com/metal-toolbox/fleetdb/pkg/api/v1"
-
-	"github.com/spf13/cobra"
-
 	mctl "github.com/metal-toolbox/mctl/cmd"
 	"github.com/metal-toolbox/mctl/internal/app"
+	"github.com/spf13/cobra"
 )
 
-// Create
+// Create.
 type createFirmwareFlags struct {
 	// file containing firmware configuration
 	firmwareConfigFile string
 }
 
-var (
-	flagsDefinedCreateFirmware *createFirmwareFlags
-)
+var flagsDefinedCreateFirmware *createFirmwareFlags
 
 var createFirmware = &cobra.Command{
 	Use:   "firmware",
@@ -55,7 +51,7 @@ var createFirmware = &cobra.Command{
 	},
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	flagsDefinedCreateFirmware = &createFirmwareFlags{}
 	usage := "JSON file with firmware configuration data"
 
diff --git a/cmd/create/firmware_set.go b/cmd/create/firmware_set.go
index 1ff386b..2e0a148 100644
--- a/cmd/create/firmware_set.go
+++ b/cmd/create/firmware_set.go
@@ -9,12 +9,11 @@ import (
 	"os"
 
 	"github.com/google/uuid"
-	"github.com/spf13/cobra"
-
 	fleetdbapi "github.com/metal-toolbox/fleetdb/pkg/api/v1"
 	mctl "github.com/metal-toolbox/mctl/cmd"
 	"github.com/metal-toolbox/mctl/internal/app"
 	"github.com/metal-toolbox/mctl/pkg/model"
+	"github.com/spf13/cobra"
 )
 
 var (
@@ -46,10 +45,7 @@ var createFirmwareSet = &cobra.Command{
 		}
 
 		if definedfirmwareSetFlags.CreateFromFile != "" {
-			err = createFWSetsFromFile(cmd.Context(), client, definedfirmwareSetFlags)
-			if err != nil {
-				log.Fatal(err)
-			}
+			createFWSetsFromFile(cmd.Context(), client, definedfirmwareSetFlags)
 		} else {
 			err = createFWSetFromCLI(cmd.Context(), client, definedfirmwareSetFlags)
 			if err != nil {
@@ -59,7 +55,7 @@ var createFirmwareSet = &cobra.Command{
 	},
 }
 
-func createFWSetsFromFile(ctx context.Context, client *fleetdbapi.Client, flgs *mctl.FirmwareSetFlags) (err error) {
+func createFWSetsFromFile(ctx context.Context, client *fleetdbapi.Client, flgs *mctl.FirmwareSetFlags) {
 	var fwsets []*fleetdbapi.ComponentFirmwareSet
 
 	fbytes, err := os.ReadFile(flgs.CreateFromFile)
@@ -72,6 +68,7 @@ func createFWSetsFromFile(ctx context.Context, client *fleetdbapi.Client, flgs *
 	}
 
 	firmwareAdded := map[string]bool{}
+
 	for _, set := range fwsets {
 		if len(set.ComponentFirmware) == 0 {
 			continue
@@ -80,7 +77,7 @@ func createFWSetsFromFile(ctx context.Context, client *fleetdbapi.Client, flgs *
 		// create firmware
 		setFwUUIDs := []string{}
 
-		// nolint:gocritic // this is fine
+		//nolint:gocritic // this is fine
 		for _, fw := range set.ComponentFirmware {
 			setFwUUIDs = append(setFwUUIDs, fw.UUID.String())
 
@@ -90,6 +87,7 @@ func createFWSetsFromFile(ctx context.Context, client *fleetdbapi.Client, flgs *
 			}
 
 			log.Printf("Adding firmware object: " + fw.UUID.String())
+
 			_, _, err = client.CreateServerComponentFirmware(ctx, fw)
 			if err != nil {
 				log.Fatal("error adding firmware object: ", err)
@@ -111,8 +109,6 @@ func createFWSetsFromFile(ctx context.Context, client *fleetdbapi.Client, flgs *
 			log.Fatal("error adding firmware-set object: ", err)
 		}
 	}
-
-	return nil
 }
 
 func createFWSetFromCLI(ctx context.Context, client *fleetdbapi.Client, flgs *mctl.FirmwareSetFlags) (err error) {
@@ -123,6 +119,7 @@ func createFWSetFromCLI(ctx context.Context, client *fleetdbapi.Client, flgs *mc
 
 	if len(definedfirmwareSetFlags.Labels) > 0 {
 		var attrs *fleetdbapi.Attributes
+
 		attrs, err = mctl.AttributeFromLabels(model.AttributeNSFirmwareSetLabels, flgs.Labels)
 		if err != nil {
 			return err
@@ -154,7 +151,7 @@ func createFWSetFromCLI(ctx context.Context, client *fleetdbapi.Client, flgs *mc
 	return nil
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	definedfirmwareSetFlags = &mctl.FirmwareSetFlags{}
 
 	mctl.AddFirmwareIDsFlag(createFirmwareSet, &definedfirmwareSetFlags.AddFirmwareUUIDs)
diff --git a/cmd/create/server.go b/cmd/create/server.go
index dd5f471..5c79b40 100644
--- a/cmd/create/server.go
+++ b/cmd/create/server.go
@@ -6,10 +6,9 @@ import (
 	"log"
 
 	coapiv1 "github.com/metal-toolbox/conditionorc/pkg/api/v1/types"
-	"github.com/spf13/cobra"
-
 	mctl "github.com/metal-toolbox/mctl/cmd"
 	"github.com/metal-toolbox/mctl/internal/app"
+	"github.com/spf13/cobra"
 )
 
 type serverEnrollParams struct {
@@ -20,9 +19,7 @@ type serverEnrollParams struct {
 	password string
 }
 
-var (
-	serverEnrollFlags *serverEnrollParams
-)
+var serverEnrollFlags *serverEnrollParams
 
 var serverEnroll = &cobra.Command{
 	Use:   "server",
@@ -67,7 +64,7 @@ func enrollServer(ctx context.Context) {
 	log.Printf("status=%d\nmsg=%s\nconditionID=%s\nserverID=%v", response.StatusCode, response.Message, condition.ID, response.Records.ServerID)
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	serverEnrollFlags = &serverEnrollParams{}
 
 	mctl.AddBMCAddressFlag(serverEnroll, &serverEnrollFlags.ip)
diff --git a/cmd/delete/delete.go b/cmd/delete/delete.go
index 325348a..a98183f 100644
--- a/cmd/delete/delete.go
+++ b/cmd/delete/delete.go
@@ -2,7 +2,6 @@ package deleteresource
 
 import (
 	"github.com/metal-toolbox/mctl/cmd"
-
 	"github.com/spf13/cobra"
 )
 
@@ -14,7 +13,7 @@ var deleteCmd = &cobra.Command{
 	},
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	cmd.RootCmd.AddCommand(deleteCmd)
 	deleteCmd.AddCommand(deleteFirmwareSet)
 	deleteCmd.AddCommand(deleteFirmware)
diff --git a/cmd/delete/firmware.go b/cmd/delete/firmware.go
index 80f6dcf..97fb13d 100644
--- a/cmd/delete/firmware.go
+++ b/cmd/delete/firmware.go
@@ -6,10 +6,9 @@ import (
 
 	"github.com/google/uuid"
 	fleetdbapi "github.com/metal-toolbox/fleetdb/pkg/api/v1"
-	"github.com/spf13/cobra"
-
 	mctl "github.com/metal-toolbox/mctl/cmd"
 	"github.com/metal-toolbox/mctl/internal/app"
+	"github.com/spf13/cobra"
 )
 
 type deleteFirmwareFlags struct {
@@ -17,9 +16,8 @@ type deleteFirmwareFlags struct {
 	id string
 }
 
-var (
-	flagsDefinedDeleteFirmware *deleteFirmwareFlags
-)
+var flagsDefinedDeleteFirmware *deleteFirmwareFlags
+
 var deleteFirmware = &cobra.Command{
 	Use:   "firmware",
 	Short: "Delete a firmware object",
@@ -45,7 +43,7 @@ var deleteFirmware = &cobra.Command{
 	},
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	flagsDefinedDeleteFirmware = &deleteFirmwareFlags{}
 
 	mctl.AddFirmwareIDFlag(deleteFirmware, &flagsDefinedDeleteFirmware.id)
diff --git a/cmd/delete/firmware_set.go b/cmd/delete/firmware_set.go
index 8e03cf2..54093f7 100644
--- a/cmd/delete/firmware_set.go
+++ b/cmd/delete/firmware_set.go
@@ -5,15 +5,12 @@ import (
 	"log"
 
 	"github.com/google/uuid"
-	"github.com/spf13/cobra"
-
 	mctl "github.com/metal-toolbox/mctl/cmd"
 	"github.com/metal-toolbox/mctl/internal/app"
+	"github.com/spf13/cobra"
 )
 
-var (
-	deleteFWSetFlags mctl.FirmwareSetFlags
-)
+var deleteFWSetFlags mctl.FirmwareSetFlags
 
 var deleteFirmwareSet = &cobra.Command{
 	Use:   "firmware-set",
@@ -40,7 +37,7 @@ var deleteFirmwareSet = &cobra.Command{
 	},
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	mctl.AddFirmwareSetFlag(deleteFirmwareSet, &deleteFWSetFlags.ID)
 	mctl.RequireFlag(deleteFirmwareSet, mctl.FirmwareSetFlag)
 }
diff --git a/cmd/delete/serverdelete.go b/cmd/delete/serverdelete.go
index c7d9671..0a0391e 100644
--- a/cmd/delete/serverdelete.go
+++ b/cmd/delete/serverdelete.go
@@ -4,20 +4,16 @@ import (
 	"context"
 	"log"
 
-	"github.com/spf13/cobra"
-
-	"github.com/metal-toolbox/mctl/internal/app"
-
 	mctl "github.com/metal-toolbox/mctl/cmd"
+	"github.com/metal-toolbox/mctl/internal/app"
+	"github.com/spf13/cobra"
 )
 
 type serverDeleteParams struct {
 	serverID string
 }
 
-var (
-	serverDeleteFlags *serverDeleteParams
-)
+var serverDeleteFlags *serverDeleteParams
 
 var serverDelete = &cobra.Command{
 	Use:   "server",
@@ -39,10 +35,11 @@ func deleteServer(ctx context.Context) {
 	if err != nil {
 		log.Fatal(err)
 	}
+
 	log.Printf("status=%d\nmsg=%s\nserverID=%v", response.StatusCode, response.Message, response.Records.ServerID)
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	serverDeleteFlags = &serverDeleteParams{}
 
 	mctl.AddServerFlag(serverDelete, &serverDeleteFlags.serverID)
diff --git a/cmd/edit/edit.go b/cmd/edit/edit.go
index 2d7eca5..fbe6f23 100644
--- a/cmd/edit/edit.go
+++ b/cmd/edit/edit.go
@@ -2,7 +2,6 @@ package edit
 
 import (
 	"github.com/metal-toolbox/mctl/cmd"
-
 	"github.com/spf13/cobra"
 )
 
@@ -14,7 +13,7 @@ var edit = &cobra.Command{
 	},
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	cmd.RootCmd.AddCommand(edit)
 	edit.AddCommand(editFirmwareSet)
 }
diff --git a/cmd/edit/firmware_set.go b/cmd/edit/firmware_set.go
index 9d49ac9..1a59511 100644
--- a/cmd/edit/firmware_set.go
+++ b/cmd/edit/firmware_set.go
@@ -6,16 +6,13 @@ import (
 
 	"github.com/google/uuid"
 	fleetdbapi "github.com/metal-toolbox/fleetdb/pkg/api/v1"
-	"github.com/spf13/cobra"
-
 	mctl "github.com/metal-toolbox/mctl/cmd"
 	"github.com/metal-toolbox/mctl/internal/app"
 	"github.com/metal-toolbox/mctl/pkg/model"
+	"github.com/spf13/cobra"
 )
 
-var (
-	editFWSetFlags mctl.FirmwareSetFlags
-)
+var editFWSetFlags mctl.FirmwareSetFlags
 
 var editFirmwareSet = &cobra.Command{
 	Use:   "firmware-set",
@@ -96,7 +93,7 @@ var editFirmwareSet = &cobra.Command{
 	},
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	mctl.AddFirmwareSetFlag(editFirmwareSet, &editFWSetFlags.ID)
 	mctl.AddNameFlag(editFirmwareSet, &editFWSetFlags.ID, "New name of the firmware set")
 	mctl.AddLabelsFlag(editFirmwareSet, &editFWSetFlags.Labels,
diff --git a/cmd/flags.go b/cmd/flags.go
index 241f145..dbe3045 100644
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -65,9 +65,7 @@ var (
 	OutputTypeText outputType = "text"
 )
 
-var (
-	errOutputType = errors.New("unsupported output type")
-)
+var errOutputType = errors.New("unsupported output type")
 
 type outputType string
 
@@ -85,13 +83,13 @@ func (o *outputType) Set(value string) error {
 	switch value {
 	case OutputTypeJSON.String(), OutputTypeText.String():
 		*o = outputType(value)
+
 		return nil
 	default:
 		return errOutputType
 	}
 }
 
-//nolint:staticcheck // SA5011 log.Fatalf will make sure we don't continue if flag is nil
 func RequireFlag(cmd *cobra.Command, flagDetail *flagDetails) {
 	flag := cmd.PersistentFlags().Lookup(flagDetail.name)
 	if flag == nil {
diff --git a/cmd/generate/docs.go b/cmd/generate/docs.go
index 5cad016..ca3efde 100644
--- a/cmd/generate/docs.go
+++ b/cmd/generate/docs.go
@@ -19,7 +19,7 @@ var cmdGenerateDocs = &cobra.Command{
 		}
 
 		linkHandler := func(s string) string { return s }
-		// nolint:revive // method is passed as a parameter
+		//nolint:revive // method is passed as a parameter
 		filePrepender := func(s string) string { return "[Auto generated by spf13/cobra]: <>\n\n" }
 		if err := cobradoc.GenMarkdownTreeCustom(cmd.Root(), "./docs", filePrepender, linkHandler); err != nil {
 			log.Fatal(err)
@@ -27,6 +27,6 @@ var cmdGenerateDocs = &cobra.Command{
 	},
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	cmd.RootCmd.AddCommand(cmdGenerateDocs)
 }
diff --git a/cmd/get/bios-config.go b/cmd/get/bios-config.go
index a608dfd..4fc5cfd 100644
--- a/cmd/get/bios-config.go
+++ b/cmd/get/bios-config.go
@@ -8,21 +8,18 @@ import (
 
 	"github.com/google/uuid"
 	fleetdbapi "github.com/metal-toolbox/fleetdb/pkg/api/v1"
-	"github.com/spf13/cobra"
-
 	mctl "github.com/metal-toolbox/mctl/cmd"
 	"github.com/metal-toolbox/mctl/internal/app"
+	"github.com/spf13/cobra"
 )
 
 type getBiosConfigFlags struct {
 	serverID string
 }
 
-var (
-	flagsDefinedGetBiosConfig *getBiosConfigFlags
-)
+var flagsDefinedGetBiosConfig *getBiosConfigFlags
 
-// Get BIOS configuration
+// Get BIOS configuration.
 var getBiosConfig = &cobra.Command{
 	Use:   "bios-config",
 	Short: "Get bios configuration information for a server",
@@ -56,7 +53,7 @@ var getBiosConfig = &cobra.Command{
 	},
 }
 
-// returns bios configuration data
+// returns bios configuration data.
 func biosConfigFromNamespaces(ctx context.Context, serverID uuid.UUID, client *fleetdbapi.Client) ([]fleetdbapi.VersionedAttributes, error) {
 	namespaces := []string{
 		"sh.hollow.alloy.inband.bios_configuration",
@@ -79,7 +76,7 @@ func biosConfigFromNamespaces(ctx context.Context, serverID uuid.UUID, client *f
 	return nil, nil
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	flagsDefinedGetBiosConfig = &getBiosConfigFlags{}
 
 	mctl.AddServerFlag(getBiosConfig, &flagsDefinedGetBiosConfig.serverID)
diff --git a/cmd/get/bom.go b/cmd/get/bom.go
index fbf762c..0dded88 100644
--- a/cmd/get/bom.go
+++ b/cmd/get/bom.go
@@ -4,10 +4,9 @@ import (
 	"log"
 
 	fleetdbapi "github.com/metal-toolbox/fleetdb/pkg/api/v1"
-	"github.com/spf13/cobra"
-
 	mctl "github.com/metal-toolbox/mctl/cmd"
 	"github.com/metal-toolbox/mctl/internal/app"
+	"github.com/spf13/cobra"
 )
 
 type getBomInfoByBmcMacAddressFlags struct {
@@ -23,9 +22,7 @@ func (gb *getBomInfoByBmcMacAddressFlags) hasBMCMacAddr() bool {
 	return gb.bmcMacAddr != ""
 }
 
-var (
-	flagsGetBomByMacAddress *getBomInfoByBmcMacAddressFlags
-)
+var flagsGetBomByMacAddress *getBomInfoByBmcMacAddressFlags
 
 var getBomInfoByMacAddress = &cobra.Command{
 	Use:   "bom",
@@ -56,7 +53,7 @@ var getBomInfoByMacAddress = &cobra.Command{
 	},
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	flagsGetBomByMacAddress = &getBomInfoByBmcMacAddressFlags{}
 
 	mctl.AddMacAOCFlag(getBomInfoByMacAddress, &flagsGetBomByMacAddress.aocMacAddr)
diff --git a/cmd/get/condition.go b/cmd/get/condition.go
index 2b96ce7..d950941 100644
--- a/cmd/get/condition.go
+++ b/cmd/get/condition.go
@@ -4,10 +4,9 @@ import (
 	"log"
 
 	"github.com/google/uuid"
-	"github.com/spf13/cobra"
-
 	mctl "github.com/metal-toolbox/mctl/cmd"
 	"github.com/metal-toolbox/mctl/internal/app"
+	"github.com/spf13/cobra"
 )
 
 type getConditionFlags struct {
@@ -15,9 +14,7 @@ type getConditionFlags struct {
 	id string
 }
 
-var (
-	flagsDefinedGetCondition *getConditionFlags
-)
+var flagsDefinedGetCondition *getConditionFlags
 
 var getCondition = &cobra.Command{
 	Use:   "condition",
@@ -44,7 +41,7 @@ var getCondition = &cobra.Command{
 	},
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	flagsDefinedGetCondition = &getConditionFlags{}
 
 	mctl.AddServerFlag(getCondition, &flagsDefinedGetCondition.id)
diff --git a/cmd/get/firmware-set.go b/cmd/get/firmware-set.go
index 3d7cd89..0456299 100644
--- a/cmd/get/firmware-set.go
+++ b/cmd/get/firmware-set.go
@@ -9,11 +9,10 @@ import (
 
 	"github.com/google/uuid"
 	fleetdbapi "github.com/metal-toolbox/fleetdb/pkg/api/v1"
-	"github.com/pkg/errors"
-	"github.com/spf13/cobra"
-
 	mctl "github.com/metal-toolbox/mctl/cmd"
 	"github.com/metal-toolbox/mctl/internal/app"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
 )
 
 type getFirmwareSetFlags struct {
@@ -21,11 +20,9 @@ type getFirmwareSetFlags struct {
 	serverID string
 }
 
-var (
-	flagsDefinedGetFirmwareSet *getFirmwareSetFlags
-)
+var flagsDefinedGetFirmwareSet *getFirmwareSetFlags
 
-// Get firmware set
+// Get firmware set.
 var getFirmwareSet = &cobra.Command{
 	Use:   "firmware-set",
 	Short: "Get information for given firmware set identifier",
@@ -110,7 +107,7 @@ func firmwareSetForServer(ctx context.Context, client *fleetdbapi.Client, server
 	return &fwSet[0], nil
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	flagsDefinedGetFirmwareSet = &getFirmwareSetFlags{}
 
 	mctl.AddServerFlag(getFirmwareSet, &flagsDefinedGetFirmwareSet.serverID)
diff --git a/cmd/get/firmware.go b/cmd/get/firmware.go
index 4691a27..07ee8ac 100644
--- a/cmd/get/firmware.go
+++ b/cmd/get/firmware.go
@@ -6,21 +6,18 @@ import (
 	"os"
 
 	"github.com/google/uuid"
-	"github.com/spf13/cobra"
-
 	mctl "github.com/metal-toolbox/mctl/cmd"
 	"github.com/metal-toolbox/mctl/internal/app"
+	"github.com/spf13/cobra"
 )
 
 type getFirmwareFlags struct {
 	id string
 }
 
-var (
-	flagsDefinedGetFirmware *getFirmwareFlags
-)
+var flagsDefinedGetFirmware *getFirmwareFlags
 
-// Get firmware info
+// Get firmware info.
 var getFirmware = &cobra.Command{
 	Use:   "firmware",
 	Short: "Get information for given firmware identifier",
@@ -50,7 +47,7 @@ var getFirmware = &cobra.Command{
 	},
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	flagsDefinedGetFirmware = &getFirmwareFlags{}
 
 	mctl.AddFirmwareIDFlag(getFirmware, &flagsDefinedGetFirmware.id)
diff --git a/cmd/get/get.go b/cmd/get/get.go
index 87da5e5..8c68ff0 100644
--- a/cmd/get/get.go
+++ b/cmd/get/get.go
@@ -1,14 +1,11 @@
 package get
 
 import (
-	"github.com/spf13/cobra"
-
 	"github.com/metal-toolbox/mctl/cmd"
+	"github.com/spf13/cobra"
 )
 
-var (
-	output string
-)
+var output string
 
 var cmdGet = &cobra.Command{
 	Use:   "get",
@@ -18,7 +15,7 @@ var cmdGet = &cobra.Command{
 	},
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	cmd.RootCmd.AddCommand(cmdGet)
 	cmdGet.AddCommand(getServer)
 	cmdGet.AddCommand(getCondition)
diff --git a/cmd/get/server.go b/cmd/get/server.go
index 1105719..453e267 100644
--- a/cmd/get/server.go
+++ b/cmd/get/server.go
@@ -10,14 +10,13 @@ import (
 	"github.com/dustin/go-humanize"
 	"github.com/google/uuid"
 	fleetdbapi "github.com/metal-toolbox/fleetdb/pkg/api/v1"
+	mctl "github.com/metal-toolbox/mctl/cmd"
+	"github.com/metal-toolbox/mctl/internal/app"
 	rfleetdb "github.com/metal-toolbox/rivets/fleetdb"
 	rt "github.com/metal-toolbox/rivets/types"
 	"github.com/olekukonko/tablewriter"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
-
-	mctl "github.com/metal-toolbox/mctl/cmd"
-	"github.com/metal-toolbox/mctl/internal/app"
 )
 
 type getServerFlags struct {
@@ -106,19 +105,21 @@ func renderServerTable(server *rt.Server, withCreds bool) {
 	tableServer.Append([]string{"Vendor", server.Vendor})
 	tableServer.Append([]string{"Serial", server.Serial})
 	tableServer.Append([]string{"BMCAddr", server.BMCAddress})
+
 	if withCreds {
 		tableServer.Append([]string{"BMCUser", server.BMCUser})
 		tableServer.Append([]string{"BMCPassword", server.BMCPassword})
 	}
+
 	tableServer.Append([]string{"Facility", server.Facility})
 	tableServer.Append([]string{"Reported", humanize.Time(server.UpdatedAt)})
-
 	tableServer.Render()
 }
 
 func renderComponentListTable(components []*rt.Component) {
 	table := tablewriter.NewWriter(os.Stdout)
 	table.SetHeader([]string{"Component", "Vendor", "Model", "Serial", "FW", "Status", "Reported"})
+
 	for _, c := range components {
 		vendor := "-"
 		model := "-"
@@ -163,8 +164,10 @@ func server(ctx context.Context, client *fleetdbapi.Client, id uuid.UUID, withCo
 	}
 
 	cserver := rfleetdb.ConvertServer(server)
+
 	if withComponents {
 		var err error
+
 		cserver.Components, err = components(ctx, client, id)
 		if err != nil {
 			return nil, err
@@ -197,7 +200,7 @@ func components(ctx context.Context, c *fleetdbapi.Client, id uuid.UUID) ([]*rt.
 	return rfleetdb.ConvertComponents(components), nil
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	cmdArgs = &getServerFlags{}
 
 	mctl.AddServerFlag(getServer, &cmdArgs.id)
diff --git a/cmd/install/firmware_set.go b/cmd/install/firmware_set.go
index 7621dbe..711fd59 100644
--- a/cmd/install/firmware_set.go
+++ b/cmd/install/firmware_set.go
@@ -6,15 +6,12 @@ import (
 	"strings"
 
 	"github.com/google/uuid"
+	fleetdbapi "github.com/metal-toolbox/fleetdb/pkg/api/v1"
+	mctl "github.com/metal-toolbox/mctl/cmd"
+	"github.com/metal-toolbox/mctl/internal/app"
+	rctypes "github.com/metal-toolbox/rivets/condition"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
-
-	"github.com/metal-toolbox/mctl/internal/app"
-
-	fleetdbapi "github.com/metal-toolbox/fleetdb/pkg/api/v1"
-	rctypes "github.com/metal-toolbox/rivets/condition"
-
-	mctl "github.com/metal-toolbox/mctl/cmd"
 )
 
 type installFirmwareSetFlags struct {
@@ -26,17 +23,14 @@ type installFirmwareSetFlags struct {
 	dryRun                bool
 }
 
-var (
-	flagsDefinedInstallFwSet *installFirmwareSetFlags
-)
+var flagsDefinedInstallFwSet *installFirmwareSetFlags
 
-// List
+// List.
 var installFirmwareSet = &cobra.Command{
 	Use:   "firmware-set",
 	Short: "Install firmware set",
 	Run: func(cmd *cobra.Command, _ []string) {
 		installFwSet(cmd.Context())
-
 	},
 }
 
@@ -129,7 +123,7 @@ func firmwareSetForInstall(ctx context.Context, client *fleetdbapi.Client, serve
 	return fwSetID, nil
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	flagsDefinedInstallFwSet = &installFirmwareSetFlags{}
 
 	mctl.AddServerFlag(installFirmwareSet, &flagsDefinedInstallFwSet.serverID)
diff --git a/cmd/install/install.go b/cmd/install/install.go
index a4e4a87..08d0792 100644
--- a/cmd/install/install.go
+++ b/cmd/install/install.go
@@ -1,9 +1,8 @@
 package install
 
 import (
-	"github.com/spf13/cobra"
-
 	"github.com/metal-toolbox/mctl/cmd"
+	"github.com/spf13/cobra"
 )
 
 var install = &cobra.Command{
@@ -14,7 +13,7 @@ var install = &cobra.Command{
 	},
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	cmd.RootCmd.AddCommand(install)
 
 	install.AddCommand(installFirmwareSet)
diff --git a/cmd/install/status.go b/cmd/install/status.go
index 5255115..9f943b5 100644
--- a/cmd/install/status.go
+++ b/cmd/install/status.go
@@ -6,11 +6,10 @@ import (
 	"log"
 
 	"github.com/google/uuid"
-	rctypes "github.com/metal-toolbox/rivets/condition"
-	"github.com/spf13/cobra"
-
 	mctl "github.com/metal-toolbox/mctl/cmd"
 	"github.com/metal-toolbox/mctl/internal/app"
+	rctypes "github.com/metal-toolbox/rivets/condition"
+	"github.com/spf13/cobra"
 )
 
 var serverIDStr string
@@ -49,7 +48,7 @@ func statusCheck(ctx context.Context) {
 	fmt.Println(s)
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	mctl.AddServerFlag(installStatus, &serverIDStr)
 	mctl.RequireFlag(installStatus, mctl.ServerFlag)
 }
diff --git a/cmd/list/component.go b/cmd/list/component.go
index 8ba178c..fe39821 100644
--- a/cmd/list/component.go
+++ b/cmd/list/component.go
@@ -5,10 +5,9 @@ import (
 	"os"
 
 	fleetdbapi "github.com/metal-toolbox/fleetdb/pkg/api/v1"
-	"github.com/spf13/cobra"
-
 	mctl "github.com/metal-toolbox/mctl/cmd"
 	"github.com/metal-toolbox/mctl/internal/app"
+	"github.com/spf13/cobra"
 )
 
 type listComponentFlags struct {
@@ -21,11 +20,9 @@ type listComponentFlags struct {
 	page    int
 }
 
-var (
-	flagsListComponent *listComponentFlags
-)
+var flagsListComponent *listComponentFlags
 
-// List
+// List.
 var listComponent = &cobra.Command{
 	Use:   "component",
 	Short: "List Components",
@@ -84,7 +81,7 @@ var listComponent = &cobra.Command{
 	},
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	flagsListComponent = &listComponentFlags{}
 
 	mctl.AddWithRecordsFlag(listComponent, &flagsListComponent.records)
diff --git a/cmd/list/firmware.go b/cmd/list/firmware.go
index 7e8b494..bfc2ac8 100644
--- a/cmd/list/firmware.go
+++ b/cmd/list/firmware.go
@@ -7,11 +7,10 @@ import (
 	"strings"
 
 	fleetdbapi "github.com/metal-toolbox/fleetdb/pkg/api/v1"
-	"github.com/olekukonko/tablewriter"
-	"github.com/spf13/cobra"
-
 	mctl "github.com/metal-toolbox/mctl/cmd"
 	"github.com/metal-toolbox/mctl/internal/app"
+	"github.com/olekukonko/tablewriter"
+	"github.com/spf13/cobra"
 )
 
 type listFirmwareFlags struct {
@@ -23,11 +22,9 @@ type listFirmwareFlags struct {
 	page      int
 }
 
-var (
-	flagsDefinedListFirmware *listFirmwareFlags
-)
+var flagsDefinedListFirmware *listFirmwareFlags
 
-// List
+// List.
 var listFirmware = &cobra.Command{
 	Use:   "firmware",
 	Short: "List firmware",
@@ -54,7 +51,7 @@ var listFirmware = &cobra.Command{
 		if flagsDefinedListFirmware.model != "" {
 			// TODO - if we really want to search using multiple models
 			//
-			//  fix the the firmware search in fleetdb, its currently useless
+			//  fix the firmware search in fleetdb, its currently useless
 			//  because fleetdb queries the data using an 'AND' instead of an 'OR'
 			filterParams.Model = []string{strings.ToLower(flagsDefinedListFirmware.model)}
 		}
@@ -89,8 +86,8 @@ var listFirmware = &cobra.Command{
 	},
 }
 
-func init() {
-	flagsDefinedListFirmware = &listFirmwareFlags{limit: 10}
+func init() { //nolint:gochecknoinits // fine
+	flagsDefinedListFirmware = &listFirmwareFlags{limit: 10} //nolint:gomnd // default is fine as is
 
 	mctl.AddVendorFlag(listFirmware, &flagsDefinedListFirmware.vendor)
 	mctl.AddModelFlag(listFirmware, &flagsDefinedListFirmware.model)
diff --git a/cmd/list/firmware_set.go b/cmd/list/firmware_set.go
index f0d3730..93a109c 100644
--- a/cmd/list/firmware_set.go
+++ b/cmd/list/firmware_set.go
@@ -6,12 +6,11 @@ import (
 	"strings"
 
 	fleetdbapi "github.com/metal-toolbox/fleetdb/pkg/api/v1"
-	"github.com/olekukonko/tablewriter"
-	"github.com/spf13/cobra"
-
 	mctl "github.com/metal-toolbox/mctl/cmd"
 	"github.com/metal-toolbox/mctl/internal/app"
 	"github.com/metal-toolbox/mctl/pkg/model"
+	"github.com/olekukonko/tablewriter"
+	"github.com/spf13/cobra"
 )
 
 type listFirmwareSetFlags struct {
@@ -19,11 +18,9 @@ type listFirmwareSetFlags struct {
 	model  string
 }
 
-var (
-	flagsDefinedListFwSet *listFirmwareSetFlags
-)
+var flagsDefinedListFwSet *listFirmwareSetFlags
 
-// List
+// List.
 var listFirmwareSet = &cobra.Command{
 	Use:   "firmware-set",
 	Short: "List firmware",
@@ -75,7 +72,7 @@ var listFirmwareSet = &cobra.Command{
 	},
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	flagsDefinedListFwSet = &listFirmwareSetFlags{}
 
 	mctl.AddModelFlag(listFirmwareSet, &flagsDefinedListFwSet.model)
diff --git a/cmd/list/list.go b/cmd/list/list.go
index 5e8188b..c4c2703 100644
--- a/cmd/list/list.go
+++ b/cmd/list/list.go
@@ -1,14 +1,11 @@
 package list
 
 import (
-	"github.com/spf13/cobra"
-
 	"github.com/metal-toolbox/mctl/cmd"
+	"github.com/spf13/cobra"
 )
 
-var (
-	output string
-)
+var output string
 
 var list = &cobra.Command{
 	Use:   "list",
@@ -18,7 +15,7 @@ var list = &cobra.Command{
 	},
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	cmd.RootCmd.AddCommand(list)
 	list.AddCommand(listFirmware)
 	list.AddCommand(listFirmwareSet)
diff --git a/cmd/list/server.go b/cmd/list/server.go
index 6730894..ba9d6fc 100644
--- a/cmd/list/server.go
+++ b/cmd/list/server.go
@@ -28,11 +28,9 @@ type listServerFlags struct {
 	page      int
 }
 
-var (
-	flagsListServer *listServerFlags
-)
+var flagsListServer *listServerFlags
 
-// List
+// List.
 var cmdListServer = &cobra.Command{
 	Use:   "server",
 	Short: "List servers",
@@ -123,6 +121,7 @@ func serversTable(servers []*rt.Server, fl *listServerFlags) {
 	}
 
 	table.SetHeader(headers)
+
 	for _, server := range servers {
 		row := []string{
 			server.ID,
@@ -195,7 +194,7 @@ func attributeParamsFromFlags(fl *listServerFlags) []fleetdbapi.AttributeListPar
 	return alp
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	flagsListServer = &listServerFlags{}
 
 	mctl.AddWithRecordsFlag(cmdListServer, &flagsListServer.records)
diff --git a/cmd/power/power.go b/cmd/power/power.go
index fa0194a..9ef492d 100644
--- a/cmd/power/power.go
+++ b/cmd/power/power.go
@@ -8,17 +8,14 @@ import (
 	"strings"
 
 	"github.com/google/uuid"
+	"github.com/metal-toolbox/conditionorc/pkg/api/v1/client"
+	coapiv1 "github.com/metal-toolbox/conditionorc/pkg/api/v1/types"
+	mctl "github.com/metal-toolbox/mctl/cmd"
+	"github.com/metal-toolbox/mctl/internal/app"
+	rctypes "github.com/metal-toolbox/rivets/condition"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"golang.org/x/exp/slices"
-
-	mctl "github.com/metal-toolbox/mctl/cmd"
-
-	"github.com/metal-toolbox/conditionorc/pkg/api/v1/client"
-	coapiv1 "github.com/metal-toolbox/conditionorc/pkg/api/v1/types"
-	rctypes "github.com/metal-toolbox/rivets/condition"
-
-	"github.com/metal-toolbox/mctl/internal/app"
 )
 
 var powerCmd = &cobra.Command{
@@ -32,7 +29,7 @@ var powerCmd = &cobra.Command{
 	},
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	mctl.RootCmd.AddCommand(powerCmd)
 }
 
@@ -72,6 +69,7 @@ func powerAction(ctx context.Context) {
 
 	if queryActionStatus {
 		actionStatus(ctx, serverID, c)
+
 		return
 	}
 
@@ -142,7 +140,7 @@ func paramsFromFlags(f *powerActionFlags) (*rctypes.ServerControlTaskParameters,
 	), nil
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	flagsDefinedPowerAction = &powerActionFlags{}
 
 	mctl.AddServerFlag(powerCmd, &flagsDefinedPowerAction.serverID)
diff --git a/cmd/root.go b/cmd/root.go
index 2d63989..ad94db9 100644
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -27,7 +27,7 @@ var (
 	reAuth  bool
 )
 
-// RootCmd represents the base command when called without any subcommands
+// RootCmd represents the base command when called without any subcommands.
 var RootCmd = &cobra.Command{
 	Use:   "mctl",
 	Short: "mctl is a CLI utility to interact with metal toolbox services",
@@ -44,7 +44,7 @@ func Execute() {
 	}
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	AddConfigFileFlag(RootCmd, &cfgFile)
 	AddReAuthFlag(RootCmd, &reAuth)
 }
diff --git a/cmd/types.go b/cmd/types.go
index 97750a8..264847e 100644
--- a/cmd/types.go
+++ b/cmd/types.go
@@ -1,6 +1,6 @@
 package cmd
 
-// firmware set command flags
+// firmware set command flags.
 type FirmwareSetFlags struct {
 	// labels are key values
 	Labels map[string]string
diff --git a/cmd/version.go b/cmd/version.go
index e300fa7..217a499 100644
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -3,9 +3,8 @@ package cmd
 import (
 	"fmt"
 
-	"github.com/spf13/cobra"
-
 	"github.com/metal-toolbox/mctl/internal/version"
+	"github.com/spf13/cobra"
 )
 
 var cmdVersion = &cobra.Command{
@@ -21,6 +20,6 @@ var cmdVersion = &cobra.Command{
 	},
 }
 
-func init() {
+func init() { //nolint:gochecknoinits // fine
 	RootCmd.AddCommand(cmdVersion)
 }
diff --git a/internal/app/app.go b/internal/app/app.go
index bb2ded7..e9e16c6 100644
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -12,12 +12,10 @@ import (
 	"github.com/spf13/viper"
 )
 
-var (
-	ErrConfig = errors.New("configuration error")
-)
+var ErrConfig = errors.New("configuration error")
 
 // Config holds configuration data when running mctl
-// App holds attributes for the mtl application
+// App holds attributes for the mtl application.
 type App struct {
 	Config *model.Config
 	// Force client to re-authenticate with Oauth services.
@@ -42,16 +40,19 @@ func openConfig(path string) (*os.File, error) {
 	if path != "" {
 		return os.Open(path)
 	}
+
 	path = viper.GetString("mctlconfig")
 	if path != "" {
 		return os.Open(path)
 	}
 
 	path = filepath.Join(xdg.Home, ".mctl.yml")
+
 	f, err := os.Open(path)
 	if err == nil {
 		return f, nil
 	}
+
 	if !errors.Is(err, os.ErrNotExist) {
 		return nil, err
 	}
@@ -65,14 +66,14 @@ func openConfig(path string) (*os.File, error) {
 }
 
 func loadConfig(cfgFile string) (*model.Config, error) {
-	cfg := &model.Config{}
 	viper.AutomaticEnv()
+
 	h, err := openConfig(cfgFile)
 	if err != nil {
 		return nil, err
 	}
 
-	cfg.File = h.Name()
+	cfg := &model.Config{File: h.Name()}
 	viper.SetConfigFile(cfg.File)
 
 	err = viper.ReadConfig(h)
@@ -88,11 +89,11 @@ func loadConfig(cfgFile string) (*model.Config, error) {
 	return cfg, nil
 }
 
-// validateClientParams checks required downstream service configuration parameters are present
+// validateClientParams checks required downstream service configuration parameters are present.
 func validateClientParams(cfg *model.Config) error {
 	if cfg.FleetDBAPI != nil {
 		if err := validateConfigOIDC(cfg.FleetDBAPI); err != nil {
-			return errors.Wrap(err, "fleetdb API API config")
+			return errors.Wrap(err, "fleetdb API config")
 		}
 	}
 
diff --git a/internal/app/clients.go b/internal/app/clients.go
index 6c80d39..2d0269c 100644
--- a/internal/app/clients.go
+++ b/internal/app/clients.go
@@ -21,7 +21,7 @@ func NewFleetDBAPIClient(ctx context.Context, cfg *model.ConfigOIDC, reauth bool
 	accessToken := "fake"
 
 	if cfg == nil {
-		return nil, errors.Wrap(ErrNilConfig, "missing fleetdb API API client configuration")
+		return nil, errors.Wrap(ErrNilConfig, "missing fleetdb API client configuration")
 	}
 
 	if cfg.Disable {
@@ -44,7 +44,7 @@ func NewFleetDBAPIClient(ctx context.Context, cfg *model.ConfigOIDC, reauth bool
 	)
 }
 
-func NewConditionsClient(ctx context.Context, cfg *model.ConfigOIDC, reauth bool) (*co.Client, error) {
+func NewConditionsClient(ctx context.Context, cfg *model.ConfigOIDC, reauth bool) (*co.Client, error) { //nolint:dupl // not actually duplicate
 	if cfg == nil {
 		return nil, errors.Wrap(ErrNilConfig, "missing conditions API client configuration")
 	}
@@ -66,7 +66,7 @@ func NewConditionsClient(ctx context.Context, cfg *model.ConfigOIDC, reauth bool
 	)
 }
 
-func NewBomServiceClient(ctx context.Context, cfg *model.ConfigOIDC, reauth bool) (*bomclient.Client, error) {
+func NewBomServiceClient(ctx context.Context, cfg *model.ConfigOIDC, reauth bool) (*bomclient.Client, error) { //nolint:dupl // not actually duplicate
 	if cfg == nil {
 		return nil, errors.Wrap(ErrNilConfig, "missing bom service API client configuration")
 	}
diff --git a/internal/auth/oauth.go b/internal/auth/oauth.go
index 5b1199b..ec9519f 100644
--- a/internal/auth/oauth.go
+++ b/internal/auth/oauth.go
@@ -27,7 +27,7 @@ const (
 
 var (
 	callbackTimeout = time.Second * 6
-	// ErrNoToken is returned when a token isn't returned from the auth flow
+	// ErrNoToken is returned when a token isn't returned from the auth flow.
 	ErrNoToken = errors.New("failed to get a token")
 )
 
@@ -57,29 +57,23 @@ func newOIDCAuthenticator(apiKind model.APIKind, cfg *model.ConfigOIDC) *authent
 func AccessToken(ctx context.Context, apiKind model.APIKind, cfg *model.ConfigOIDC, reauth bool) (string, error) {
 	authenticator := newOIDCAuthenticator(apiKind, cfg)
 
-	var token *oauth2.Token
-	var err error
-
-	if reauth {
-		token, err = authenticator.getOAuth2Token(ctx)
-		if err != nil {
-			return "", err
-		}
-	} else {
-		token, err = authenticator.refreshToken(ctx)
-		if err != nil {
-			token, err = authenticator.getOAuth2Token(ctx)
-			if err != nil {
-				return "", err
-			}
-		}
+	token, err := authenticator.getOAuth2Token(ctx, !reauth)
+	if err != nil {
+		return "", err
 	}
 
 	return token.AccessToken, nil
 }
 
 // GetOAuth2Token retrieves the OAuth2 token from the issuer and stores it in the local keyring with the given name.
-func (a *authenticator) getOAuth2Token(ctx context.Context) (*oauth2.Token, error) {
+func (a *authenticator) getOAuth2Token(ctx context.Context, refresh bool) (*oauth2.Token, error) {
+	if refresh {
+		token, err := a.refreshToken(ctx)
+		if err == nil {
+			return token, nil
+		}
+	}
+
 	oauthConfig, err := a.oauth2Config(ctx)
 	if err != nil {
 		return nil, err
@@ -159,7 +153,7 @@ func (a *authenticator) refreshToken(ctx context.Context) (*oauth2.Token, error)
 	return newToken, nil
 }
 
-// tokenFromRaw will take a access and refresh token string and convert them into a proper token
+// tokenFromRaw will take a access and refresh token string and convert them into a proper token.
 func (a *authenticator) tokenFromRaw(rawAccess, refresh string) (*oauth2.Token, error) {
 	tok, err := jwt.ParseSigned(rawAccess)
 	if err != nil {
@@ -189,11 +183,11 @@ func (a *authenticator) keyringStoreToken(token *oauth2.Token) error {
 }
 
 // authCodePKCE starts a server and listens for an oauth2 callback and will
-// return the API token to the caller
+// return the API token to the caller.
 func (a *authenticator) authCodePKCE(ctx context.Context, oauthConfig *oauth2.Config, audience string) (*oauth2.Token, error) {
 	tc := make(chan *oauth2.Token)
 
-	// nolint:gomnd // state string is limited to 20 random characters
+	//nolint:gomnd // state string is limited to 20 random characters
 	c := &authClient{
 		oauthConfig: oauthConfig,
 		state:       randStr(20),
@@ -206,7 +200,7 @@ func (a *authenticator) authCodePKCE(ctx context.Context, oauthConfig *oauth2.Co
 		c.handlePKCECallback(ctx, w, r, tc)
 	})
 
-	// nolint:gomnd // read header timeout is set to 30s
+	//nolint:gomnd // read header timeout is set to 30s
 	server := &http.Server{Addr: ":18000", ReadHeaderTimeout: time.Second * 30, Handler: mux}
 
 	go func() {
@@ -275,7 +269,6 @@ func (c *authClient) handlePKCECallback(ctx context.Context, w http.ResponseWrit
 	token, err := c.oauthConfig.Exchange(ctx, code,
 		oauth2.SetAuthURLParam("code_verifier", c.codeVerifier.String()),
 	)
-
 	if err != nil {
 		log.Printf("ERROR in token exchange: %s\n", err.Error())
 
diff --git a/pkg/model/model.go b/pkg/model/model.go
index 55e9a0a..56974ad 100644
--- a/pkg/model/model.go
+++ b/pkg/model/model.go
@@ -14,9 +14,8 @@ const (
 	BomsServiceAPI APIKind = "bomservice"
 )
 
-// Config struct holds the mctl configuration parameters
+// Config struct holds the mctl configuration parameters.
 type Config struct {
-
 	// File is configuration file path
 	File       string
 	FleetDBAPI *ConfigOIDC `mapstructure:"serverservice_api"` // TODO: implement backwards compatibility and rename.
```
</details>

DoctorVin:
> I'd like to make this config be the new source-of-truth and would be happy to make sure it works in mctl too. My favorite part of this config is that we explicitly opt out of linters. This is nice because when we do update to a newer golangci-lint that has new linters we'd be in the right time to evaluate adopting or not. This also has the benefit of actually using the linters we think we are using. The whole reason I messed around with the config here is because `gofumpt` wasn't actually enabled and when I ran it as part of my editor's auto-save hook the changes were unexpected!

I have some strong opinions as to what linters are actually useful, and those biases are reflected in mctl and condition-orc most. I'm pretty against forcing us to opt-out of new linters and would rather we curate a list. Opting out explicitly means that it's not easy to answer the question of "what specific linters are we using?"

mmlb:
> > I'd like to make this config be the new source-of-truth and would be happy to make sure it works in mctl too. My favorite part of this config is that we explicitly opt out of linters. This is nice because when we do update to a newer golangci-lint that has new linters we'd be in the right time to evaluate adopting or not. This also has the benefit of actually using the linters we think we are using. The whole reason I messed around with the config here is because `gofumpt` wasn't actually enabled and when I ran it as part of my editor's auto-save hook the changes were unexpected!
> 
> I have some strong opinions as to what linters are actually useful, and those biases are reflected in mctl and condition-orc most. I'm pretty against forcing us to opt-out of new linters and would rather we curate a list. Opting out explicitly means that it's not easy to answer the question of "what specific linters are we using?"

Yep I agree with a lot of what you've said, but by doing a whitelist we miss out on potentially useful new linters. Opting out makes sure we keep the curated list up to date. I don't think the opt-out is too bad if we're pinning golangci-lint to specific versions since it won't pick up new linters underneath our feet. Using disable over enable also gives the opportunity to comment why a linter is disabled which is beneficial too.

A useful example would be tenv and thelper. They aren't enabled now and I've disabled them in this PR. There's a comment saying we should enable it and I definitely think we should, thelper points out one case that is missing t.Helper() call and tenv is easier to use than managing the envs ourselves and avoids accidentally influencing other tests.

mmlb:
Also by opting out you can make your strong opinions known and with rationale instead of letting them just be implicit. Are tenv/thelper not enbaled because you don't like them or because you weren't aware of them?

DoctorVin:
> Yep I agree with a lot of what you've said, but by doing a whitelist we miss out on potentially useful new linters. Opting out makes sure we keep the curated list up to date. I don't think the opt-out is too bad if we're pinning golangci-lint to specific versions since it won't pick up new linters underneath our feet. Using disable over enable also gives the opportunity to comment why a linter is disabled which is beneficial too.

That's a good point. I am most against picking up whatever new code style is this-week's fashion. However, because of the way that golangci-lint works we often have to update that on the fly because the golang version changed and their internal `typecheck` test starts failing in weird ways. I guess my point is: it's not always a thoughtful, planned change to golangci-lint. Updating the engine and grabbing additional linters that need disabling can be an unwelcome surprise.
 
> A useful example would be tenv and thelper. They aren't enabled now and I've disabled them in this PR. There's a comment saying we should enable it and I definitely think we should, thelper points out one case that is missing t.Helper() call and tenv is easier to use than managing the envs ourselves and avoids accidentally influencing other tests.

Again, this is a good point. I'm not opposed to adding in new linters, I just find a lot of them (looking specifically at `wsl` and `go-mnd`) have very limited utility.

DoctorVin:
> ACK but also could be useful especially when updating go versions (backwards compatibility is great in go but sometimes forwards not as great, surprises are definitely expected sometimes).

OK. I'm willing to try it your way for a while. It might work fine, so let's find out.

> wsl and gomnd have been the most annoying ones indeed! I only addressed them because I saw some `//nolint:wsd` and some `wsd` in the config so figured they were wanted.
😱 heck no!

mmlb:
What about the separate repo idea? Either that or a comment in the config file showing where it came from (or both) make sense to me.

DoctorVin:
Pro: I like single-sources of truth.Con: I really hate `git submodules` and `git subtree` is somewhat better but still not awesome.A comment in the config to a repo and a commit might be the least ugly way to deal with it? On Apr 11, 2024, at 3:15 PM, Manuel Mendez ***@***.***> wrote:﻿
What about the separate repo idea? Either that or a comment in the config file showing where it came from (or both) make sense to me.

mmlb:
https://github.com/metal-toolbox/golangci-lint-config/pull/1

mmlb:
Alright @joelrebel, @DoctorVin I've updated the config file to be based off of mctl's. There were some invalid configs, cargo culted settings and unnecessary disables that I've removed/tweaked. Both mctl and ironlib will need some tweaks to pass with this config but imo they are worth doing. mctl is pretty good, ironlib needs the most work (mostly mechanical). I'll post a PR to each with the changes necessary to work with this config in a little bit.

mmlb:
It would probably be easier to review each commit on its own, I made it so it would be easier to trust the changes/diff if reviewed that way.

actions/inventory.go
```diff
@@ -903,28 +904,54 @@ func (a *InventoryCollectorAction) acceptChange(change *diff.Change) bool {
 func (a *InventoryCollectorAction) vetUpdate(change *diff.Change) bool {
 	// allow vendor, model field changes only if the older value was not defined
 	if slices.Contains(change.Path, "Vendor") || slices.Contains(change.Path, "Model") {
-		if strings.TrimSpace(change.To.(string)) != "" && strings.TrimSpace(change.From.(string)) == "" {
+		from, ok := change.From.(string)
```
joelrebel, line 907: thanks for these validations

.github/workflows/push-pr-lint.yaml
```diff
@@ -8,7 +8,7 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v4
       with:
-        go-version: '1.19'
```
joelrebel: the `go-version-file: go.mod` would be nicer, so we don't have to keep bumping this value

joelrebel (commented):
Thanks for fixing the context passing and all the linting efforts. 
CodeQL requires some fixing similar to  https://github.com/metal-toolbox/mctl/blob/9d20393342f72e2d12083d92bf8c32e861b22038/.github/workflows/codeql-analysis.yml#L35
https://github.com/metal-toolbox/mctl/blob/9d20393342f72e2d12083d92bf8c32e861b22038/.github/workflows/codeql-analysis.yml#L43

.github/workflows/push-pr-lint.yaml
mmlb: oh yeah I didn't know about that

mmlb:
> Thanks for fixing the context passing and all the linting efforts. CodeQL requires some fixing similar to [metal-toolbox/mctl@`9d20393`/.github/workflows/codeql-analysis.yml#L35](https://github.com/metal-toolbox/mctl/blob/9d20393342f72e2d12083d92bf8c32e861b22038/.github/workflows/codeql-analysis.yml#L35) [metal-toolbox/mctl@`9d20393`/.github/workflows/codeql-analysis.yml#L43](https://github.com/metal-toolbox/mctl/blob/9d20393342f72e2d12083d92bf8c32e861b22038/.github/workflows/codeql-analysis.yml#L43)

Will address. Do you know what we need to do so the codesign check accepts the PR merge commit from GH?

mmlb:
I've updated to address feedback, ptal @joelrebel

joelrebel (approved):
🚀

mmlb:
Lets do it!

mmlb:
squash and merge make me sad :(

turegano-equinix:
> squash and merge make me sad :(

and me xD

```
This branch has conflicts that must be resolved
Use the [web editor](https://github.com/metal-toolbox/ironlib/pull/131/conflicts) or the to resolve conflicts.
Conflicting files
.golangci.yml
```

PR #133, opened 2024-04-22
House Cleaning Part 2 - Electric Boogaloo!

mmlb:
### What does this PR do

Gets rid of //nolint directives that were no longer needed or trivially fixed. Should have done this in the last PR but I forgot. Also fixes the CODEOWNERS file to be valid, moved it into .github dir and added myself and @turegano-equinix to it.


actions/inventory.go
```diff
@@ -341,14 +341,11 @@ func (a *InventoryCollectorAction) setDefaultAttributes() {
 
 // CollectDrives executes drive collectors and merges the data into device.[]*Drive
 // nolint:gocyclo // TODO(joel) if theres more conditionals to be added in here, the method is to be split up.
-func (a *InventoryCollectorAction) CollectDrives(ctx context.Context) error {
-	// nolint:errcheck // deferred method catches a panic, error check not required.
-	defer func() error {
+func (a *InventoryCollectorAction) CollectDrives(ctx context.Context) (err error) {
```
mmlb, line 344: I wanted to make sure you saw these @joelrebel. These are the only changes that potentially impact runtime. This is a change in behavior because now `a.failOnError` is actually going to influence the returned error. My first reaction to realizing that this code was wrong was to just do `defer func() { _ = recover() }()` but then thought maybe I should do what was intended instead. Would love your thoughts here.

mmlb:
I left you as an OWNER @diogomatsubara let me know if thats not useful/appropriate though.

.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @diogomatsubara @joelrebel @mmlb @turegano-equinix
```
joelrebel, line 1: Could you drop Diogo from this list and include @splaspood

actions/inventory.go
joelrebel: When you first pointed this out, I thought the initial code was incorrectly overwritting the returned error. 
But it seems like the `return nil` in the initial code defer func would not overwrite the returned error.

I tried both variants and they seem to behave the same https://go.dev/play/p/ktvkitbWyh5
so the change looks fine to me, let me know if Im missing something obvious.

utils/dsu_test.go
```diff
@@ -86,10 +86,7 @@ func Test_dsuParsePreviewBytes(t *testing.T) {
 func Test_findDSUInventoryCollector(t *testing.T) {
 	invb := "invcol_5N2WM_LN64_20_09_200_921_A00.BIN"
 
-	tmpDir, err := os.MkdirTemp("/tmp", "ironlibtest")
-	if err != nil {
-		t.Error(err)
-	}
+	tmpDir := t.TempDir()
```
joelrebel, line 89: TIL: t.TempDir() is a thing

joelrebel (commented):
Thanks for this cleanup, left a few minor suggestions

utils/dsu_test.go
mmlb: test helpers in stdlib always catch my eye

actions/inventory.go
mmlb: Yep it seems to me that returns from defer'd functions are just silently dropped

.github/CODEOWNERS
mmlb: done

mmlb:
@joelrebel I've updated the CODEOWNERS file, I believe that was it right? PTAL.

joelrebel (approved):

PR #135, opened 2024-04-23
ironlib is able to detect ineffective wipes

turegano-equinix:
### What does this PR do
Disks lie about supporting standard commands and/or successful completion of those commands, so we need to “trust but verify”. We have definitely seen this before, most recently with some micron nvme disks that respond successfully to TRIM commands but do not actually wipe the data

Added in examples/diskwipe/main.go

ironlib is able to detect ineffective wipes

### DIFF
--- a/actions/storage_controller.go
+++ b/actions/storage_controller.go
@@ -102,8 +102,10 @@ func (s *StorageControllerAction) WipeDisk(ctx context.Context, logicalName stri
 	}
 	// Watermark disk
 	log.Printf("%s | Initiating watermarking process", logicalName)
-	check := utils.ApplyWatermarks(logicalName)
-
+	check, err := utils.ApplyWatermarks(logicalName)
+	if err != nil {
+		return err
+	}
 	err = util.WipeDisk(ctx, logicalName)
 	if err != nil {
 		return err
--- a/actions/storage_controller.go
+++ b/actions/storage_controller.go
@@ -13,10 +13,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-var (
-	ErrVirtualDiskManagerUtilNotIdentified = errors.New("virtual disk management utility not identifed")
-	ErrWipeVerificationNotSatisfactory     = errors.New("ineffective wipe")
-)
+var ErrVirtualDiskManagerUtilNotIdentified = errors.New("virtual disk management utility not identifed")
 
 type StorageControllerAction struct {
 	Logger *logrus.Logger
@@ -104,18 +101,18 @@ func (s *StorageControllerAction) WipeDisk(ctx context.Context, logicalName stri
 		return err
 	}
 	// Watermark disk
-	check, err := utils.PrepWatermarks(logicalName)
-	if err != nil {
-		return err
-	}
+	log.Printf("%s | Initiating watermarking process", logicalName)
+	check := utils.ApplyWatermarks(logicalName)
 
 	err = util.WipeDisk(ctx, logicalName)
 	if err != nil {
 		return err
 	}
-
-	if check() {
-		return nil
+	log.Printf("%s | Checking if the watermark has been removed", logicalName)
+	err = check()
+	if err != nil {
+		return err
 	}
-	return errors.Wrap(ErrWipeVerificationNotSatisfactory, "Watermark still present, disk not wipped")
+	log.Printf("%s | Watermarks has been removed", logicalName)
+	return nil
 }
--- a/actions/storage_controller.go
+++ b/actions/storage_controller.go
@@ -13,7 +13,10 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-var ErrVirtualDiskManagerUtilNotIdentified = errors.New("virtual disk management utility not identifed")
+var (
+	ErrVirtualDiskManagerUtilNotIdentified = errors.New("virtual disk management utility not identifed")
+	ErrWipeVerificationNotSatisfactory     = errors.New("ineffective wipe")
+)
 
 type StorageControllerAction struct {
 	Logger *logrus.Logger
@@ -100,6 +103,19 @@ func (s *StorageControllerAction) WipeDisk(ctx context.Context, logicalName stri
 	if err != nil {
 		return err
 	}
+	// Watermark disk
+	check, err := utils.PrepWatermarks(logicalName)
+	if err != nil {
+		return err
+	}
 
-	return util.WipeDisk(ctx, logicalName)
+	err = util.WipeDisk(ctx, logicalName)
+	if err != nil {
+		return err
+	}
+
+	if check() {
+		return nil
+	}
+	return errors.Wrap(ErrWipeVerificationNotSatisfactory, "Watermark still present, disk not wipped")
 }
--- a/actions/storage_controller.go
+++ b/actions/storage_controller.go
@@ -101,20 +101,24 @@ func (s *StorageControllerAction) WipeDisk(ctx context.Context, logicalName stri
 		return err
 	}
 	// Watermark disk
+	// Before wiping the disk, we apply watermarks to later verify successful deletion
 	log.Printf("%s | Initiating watermarking process", logicalName)
 	check, err := utils.ApplyWatermarks(logicalName)
 	if err != nil {
 		return err
 	}
+	// Wipe the disk
 	err = util.WipeDisk(ctx, logicalName)
 	if err != nil {
 		return err
 	}
+	// Check if the watermark has been removed after wiping
 	log.Printf("%s | Checking if the watermark has been removed", logicalName)
 	err = check()
 	if err != nil {
 		return err
 	}
+	// Watermarks have been successfully removed, indicating successful deletion
 	log.Printf("%s | Watermarks has been removed", logicalName)
 	return nil
 }
--- a/utils/watermark_disk.go
+++ b/utils/watermark_disk.go
@@ -21,30 +21,23 @@ type watermark struct {
 	data     []byte
 }
 
-func ApplyWatermarks(logicalName string) func() error {
+func ApplyWatermarks(logicalName string) (func() error, error) {
 	// Write open
 	file, err := os.OpenFile(logicalName, os.O_WRONLY, 0)
 	if err != nil {
-		return func() error {
-			return err
-		}
+		return nil, err
 	}
 	defer file.Close()
 
 	// Get disk or partition size
 	fileSize, err := file.Seek(0, io.SeekEnd)
 	if err != nil {
-		return func() error {
-			return err
-		}
+		return nil, err
 	}
 	// Write watermarks on random locations
 	watermarks := writeWatermarks(file, 0, fileSize, numWatermarks)
 	if len(watermarks) != numWatermarks {
-		return func() error {
-			ErrorWritingWatermarks := errors.New("Error writing watermarks in the file")
-			return fmt.Errorf("%s | %w", logicalName, ErrorWritingWatermarks)
-		}
+		return nil, err
 	}
 
 	checker := func() error {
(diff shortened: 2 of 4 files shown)


utils/watermark_disk.go
```diff
@@ -0,0 +1,105 @@
...
+		_, err = file.Seek(int64(randomPosition.Uint64()), io.SeekStart)
+		if err != nil {
+			log.Println("Error moving rw pointer:", err)
+			return err
+		}
+		watermarks[i].position = randomPosition.Uint64()
+		watermarks[i].data = data
+	}
+	return nil
+}
```
mmlb, line 105:
```suggestion
func writeWatermarks(r *mathrand.Rand, dev io.WriteSeeker, count int, bufferSize int64) ([]watermarks, error) {
	devSize, err := dev.Seek(0, io.SeekEnd)
	if err != nil {
		return nil, err
	}

	partitionSize := int64(devSize / count)
	marks := make([]watermarks, 0, count)

	// Divide the device into count "partitions"
	// Then we write random data somewhere randomly within each partition
	// This way we spread out the writes through out the disk and avoid any groupings (random is random after all)
	for whence := int64(0); whence < devSize; whence += partitionSize {
		data := make([]byte, bufferSize)
		_, err := rand.Read(data)
		if err != nil {
			return nil, err
		}

		offset, err := r.Int64N(partitionSize - bufferSize)
		if err != nil {
			return nil, err
		}

		pos := whence + offset
		_, err = dev.Seek(pos, io.SeekStart)
		if err != nil {
			return nil, err
		}

		err = dev.Write(data)
		if err != nil {
			return nil, err
		}
		watermarks = append(watermarks, watermark{position: pos, data: data})
	}
	return nil
}
```

I've reworked this function with a couple of enhancements.

This new signature should be more amenable for testing, which we should definitely add for this functionality :D.

The other thing it does is makes sure to divide the disk into count chunks/partitions so that we are sure to spread the watermarks all throughout the disks. Since the previous code was just getting random positions all over the disks there could be a chance that we were writing to only a small section of the disk, or even having overlapping writes. I also tweaked the random position calculation to make sure we have space for the full write w/o crossing partitions or the end-of-disk.

mmlb: Also not a big deal but maybe worth discussing. Instead of saving the data we can save a sha256 or w/e hash of the data instead. This way we can easily grow the data size without having to worry about memory usage.

mmlb (changes requested):

utils/watermark_disk.go
mmlb: I was thinking more bytes not more marks and we will be running this in parallel against each disk but yeah not a big deal. 

https://pkg.go.dev/math/rand/v2#Rand.Int64N

utils/watermark_disk.go
```diff
@@ -0,0 +1,114 @@
...
+	bufferSize    = 512
+	numWatermarks = 10
+)
+
+type watermark struct {
+	position int64
+	data     []byte
+}
+
+func ApplyWatermarks(logicalName string) func() error {
```
mmlb, line 24:
```suggestion
func ApplyWatermarks(logicalName string) (func() error, error) {
```
Sorry if I misled you before but I meant that cleanup func should return error not bool, but we should still return errors in this function too for the errors that are hit in this func

utils/watermark_disk.go
```diff
@@ -0,0 +1,114 @@
...
+		return []watermark{w}
+	}
+	// Divide the intervals into two equal parts (approximately)
+	mid := (a + b) / 2
+
+	// Return recursively the call of function with the two remaining intervals
+	leftCount := count / 2
+	rightCount := count - leftCount
+
+	return append(writeWatermarks(file, a, mid-1, leftCount), writeWatermarks(file, mid, b, rightCount)...)
```
mmlb, line 113: why recurse instead of loop?

turegano-equinix: It came to my mind as a divide and conquer algorithm, it breaks the problem info into smaller ones, and you only need to implement the base case and split otherwise

https://upload.wikimedia.org/wikipedia/commons/e/e6/Merge_sort_algorithm_diagram.svg

mmlb: seems unnecessary and confusing to me. loop > recursion for readability most times imo :D. The function signature also shows that, what is a and whit is b?

turegano-equinix: in math you define a closed interval with [a,b] in our case [0, fileSize]. if watermarks > 1 you split the interval in two [0, mid -1] and [mid, fileSize]. each interval is now smaller and the watermark count decreases for each call. I feel that using recursion here is an elegant and concise way to solve it. I'm not agree with loop > recursion for example fibonacci sequence

turegano-equinix: Changed to iterative! Thx

utils/watermark_disk.go
```diff
@@ -0,0 +1,101 @@
...
+			// Read the watermark written to the position
+			currentValue := make([]byte, bufferSize)
+			_, err = io.ReadFull(file, currentValue)
+			if err != nil {
+				return err
+			}
+			// Check if the watermark is still in the disk
+			if slices.Equal(currentValue, watermark.data) {
+				ErrorExistingWatermark := errors.New("Error existing watermark in the file")
+				return fmt.Errorf("%s | %w", logicalName, ErrorExistingWatermark)
```
mmlb, line 64: it would be good to know the position maybe?

mmlb (changes requested):
Would love to see a test for this. Both postive and negative tests. Negative could be create a ~15KB empty file, and run check func after this wipe. It should error out.

Then for the postive case: wipe the same file (truncate to 0b then seek to a position is quick and easy, let the kernel figure it out :D) and run check again it should pass.

Maybe write the file full of random data and run check again to make sure it still comes back with an error? I don't really think this last one is necessary though.

utils/watermark_disk.go
turegano-equinix: implemented and the tests

utils/watermark_disk_test.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+		assert.ErrorContains(t, err, "invalid argument")
+	})
+
+	t.Run("EmptyFile", func(t *testing.T) {
+		// Wipe the file
+		err := os.WriteFile(tempFile.Name(), make([]byte, 0), 0o600)
+		if err != nil {
+			t.Fatalf("Failed to truncate file: %v", err)
+		}
```
mmlb, line 46:
```suggestion
		assert.NoError(t, os.Truncate(tempfile.Name(), 0))
```

utils/watermark_disk_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// Close the file since we'll be reopening it in ApplyWatermarks
+	tempFile.Close()
+
+	t.Run("NegativeTest", func(t *testing.T) {
+		// Create a ~15KB empty file, no room for all watermarks
+		err := os.WriteFile(tempFile.Name(), make([]byte, 1*1024), 0o600)
+		if err != nil {
+			t.Fatalf("Failed to create empty file: %v", err)
+		}
```
mmlb, line 27:
```suggestion
		// Create a 1KB empty file, no room for all watermarks
		assert.NoError(t, os.Truncate(tempFile.Name(), make(1*1024)))
```

utils/watermark_disk_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+		// Apply watermarks and expect an error
+		checker, err := ApplyWatermarks(tempFile.Name())
+
+		if checker != nil {
+			err = checker()
+			if err == nil {
+				t.Error("Expected error, got none")
+			}
+		}
```
mmlb, line 37:
```suggestion
		assert.Nil(t, checker)
```
we should not return a func if there's an error

utils/watermark_disk_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+		// Apply watermarks and expect no error
+		checker, err := ApplyWatermarks(tempFile.Name())
+		assert.Error(t, err, "No space for watermarking")
+		if checker != nil {
+			err := checker()
+			if err != nil {
+				t.Errorf("Expected no error, got: %v", err)
+			}
+		}
```
mmlb, line 56:
```suggestion
		assert.Nil(t, checker)
```

utils/watermark_disk_test.go
```diff
@@ -0,0 +1,87 @@
...
+		checker, err := ApplyWatermarks(tempFile.Name())
+		if err != nil {
+			t.Fatalf("Error applying watermarks: %v", err)
+		}
+		// simulate wipe
+		zeroData := make([]byte, 15*1024*1024)
+		err = os.WriteFile(tempFile.Name(), zeroData, 0o600)
+		if err != nil {
+			t.Fatalf("Failed to write zero data to file: %v", err)
+		}
```
mmlb, line 80: I like the following better since its less code but whats in here now is ok too:
```suggestion
		assert.NoError(t, os.Truncate(tempFile.Name(), 0))
		assert.NoError(t, os.Truncate(tempFile.Name(), 15*1024*1024))
```

utils/watermark_disk_test.go
```diff
@@ -0,0 +1,87 @@
+package utils
+
+import (
+	"crypto/rand"
+	"os"
+	"testing"
+
+	"gotest.tools/assert"
```
mmlb, line 8: hmm most of the other tests here use `"github.com/stretchr/testify/assert"` looks like only providers/dell/dell_test.go is using `"gotest.tools/assert"` we should probably just use one only.

utils/watermark_disk.go
```diff
@@ -60,8 +65,8 @@ func ApplyWatermarks(logicalName string) (func() error, error) {
 			}
 			// Check if the watermark is still in the disk
 			if slices.Equal(currentValue, watermark.data) {
-				ErrorExistingWatermark := errors.New("Error existing watermark in the file")
-				return fmt.Errorf("%s | %w", logicalName, ErrorExistingWatermark)
+				ErrorExistingWatermark := errors.New("Error existing watermark in the position: ")
+				return fmt.Errorf("%s | %w %d", logicalName, ErrorExistingWatermark, watermark.position)
```
mmlb, line 69: maybe `"%s@%d | %w"` instead? Otherwise the location is going to be after the error string when it should be closer to the device name I think.

mmlb (changes requested):

utils/watermark_disk_test.go
joelrebel: +1 lets stick to `testify/assert`, the package in dell_test.go was most likely a mistake

utils/watermark_disk.go
```diff
@@ -0,0 +1,106 @@
...
+	bufferSize    = 512
+	numWatermarks = 10
+)
+
+type watermark struct {
+	position int64
+	data     []byte
+}
+
+func ApplyWatermarks(logicalName string) (func() error, error) {
```
joelrebel, line 24: Include a comment to explain a bit more about this function

actions/storage_controller.go
```diff
@@ -100,6 +100,21 @@ func (s *StorageControllerAction) WipeDisk(ctx context.Context, logicalName stri
 	if err != nil {
 		return err
 	}
-
-	return util.WipeDisk(ctx, logicalName)
+	// Watermark disk
```
joelrebel, line 103: Why is the watermark required before wipe, include a comment :)

utils/watermark_disk_test.go
turegano-equinix: This is what you meant, right?

`assert.NoError(t, os.Truncate(tempFile.Name(), 1*1024))`

utils/watermark_disk_test.go
turegano-equinix: Less code is better code ;)

utils/watermark_disk_test.go
turegano-equinix: assert := `github.com/stretchr/testify/assert`

joelrebel (approved):

PR #137, opened 2024-04-29
Quiet down tests

mmlb:
This PR fixes a few things wrong with the logging setup.

First it adds logrus as an arg to the WipeDisk interface method. It makes sense to have progress updates during WipeDisk since its possible the wipe will take a long time and callers want progress info (maybe we make progress printing a callback instead of logging). Previously the code was just using stdlib's log package for logging which doesn't go with the rest of the project and doesn't allow for callers to tweak the output. Also log lines end up in test output unnecessarily.

While messing around here I noticed that many structs had a `trace` parameter that was mostly ignored, instead opting to calculate the bool value every time a function needed it. So I changed the code to set and use s.trace instead. Maybe a better idea will be to pass Executors into utils that already has trace/no-trace and envvars too.

Next I fixed up providers/dell to use stretchr/testify instead gotools.assert this was the only place where this dep was used and was likely a mistake.

Finally I changed `logrus.New` calls in test files to use `test.NewNullLogger()` so that test output won't be printed unnecessarily (see commit message for example).


mmlb:
I like passing the executor down because we already need that with the Fake$util and Fake$executor we already do.

joelrebel:
Passing the executor in requires the caller to initialise the executor (and not reuse) before invoking a utility.
I like that currently each utility currently self contained and is capable of initialising itself,
How do you see this working in practice though?

mmlb:
Yeah that’s true but it seems like something that should be left to main no? Similar to logging? I can’t quite put my finger on it but it seems like trace, the LC override and fake executors are hinting that this should be a parameter (though maybe not the LC override, I get why that’s being done). Passing in executors would possibly make the mocking simpler too if we don’t need to have one giant fake executor.

joelrebel:
> Yeah that’s true but it seems like something that should be left to main no? Similar to logging? I can’t quite put my finger on it but it seems like trace, the LC override and fake executors are hinting that this should be a parameter (though maybe not the LC override, I get why that’s being done). Passing in executors would possibly make the mocking simpler too if we don’t need to have one giant fake executor.

How about each utility constructor method optionally accepts - a logger, executor and anything else as variadic parameters based on the [options pattern](https://github.com/tmrts/go-patterns/blob/master/idiom/functional-options.md). And if the Executor isn't provided, then the constructor initialises it.

mmlb:
> > Yeah that’s true but it seems like something that should be left to main no? Similar to logging? I can’t quite put my finger on it but it seems like trace, the LC override and fake executors are hinting that this should be a parameter (though maybe not the LC override, I get why that’s being done). Passing in executors would possibly make the mocking simpler too if we don’t need to have one giant fake executor.
> 
> How about each utility constructor method optionally accepts - a logger, executor and anything else as variadic parameters based on the [options pattern](https://github.com/tmrts/go-patterns/blob/master/idiom/functional-options.md?rgh-link-date=2024-04-30T16%3A21%3A33Z). And if the Executor isn't provided, then the constructor initialises it.

I thought of that but didn't mention it because ideally we'd always pass these in imo. The utilities shouldn't be logging using stdlib's log so should definitely get this as a required param (or maybe not log if the option wasn't passed in?) (I think I missed one or 2 utils that are logging when they shouldn't be, need to check and update if so.) I think having Executor be used similarly to context/logger would be nice. Have one "main" Executor object and be able to SetEnv, SetArgs, SetQuiet/Verbose returns a new Executor.

joelrebel:
Im skeptical of moving the responsibility of initialising the executor to the caller completely, imo an Options pattern would work better and prevent the caller from shooting themselves in the foot (initialize a smartctl executor and pass that to nvmecli for example). 

Would be worth having some thoughts from @DoctorVin and @splaspood in this case.

DoctorVin:
Joel's point about someone making a mistake and running the same utility in a different executor seems on-target in terms of avoiding really difficult-to-debug errors. I too favor the options pattern where we can apply it cleanly. Right now this library seems pretty wedded to the idea of having bespoke constructors; because of that I think the options pattern makes the most sense. If we had more a generic factory-style instantiation for the utils my opinion might be different.

mmlb:
I'd like to put the different `Executor` discussion on hold, I shouldn't have brought it up without concrete examples or fixes. And I don't intend to do that sort of change in this PR. I do intend to write up a PR (or 2, not sure yet :smile:) to demonstrate the friction I see and what potential solutions will look like, for discussion.
 
That being said, I'd like to get this PR reviewed in its current state so I can get the new functionality unblocked and stop confusing myself will all the PRs :sweat_smile:.

joelrebel (approved):
Current changes look good to me 👍

PR #129, opened 2024-04-11
Better nvme capability detection

mmlb:

PR #141, opened 2024-05-01
More refactors, fixes, etc

mmlb:
I've split these commits out from #136 where they came up. The commits here are grouped into fixes and qol stuff. The QoL stuff are changes to Executor and some missed code comments/fixes to nvme that were meant to go in during #129.

The other big chunk of changes is to utils/watermark where I ran into incorrect params to rand.Int that would cause us to seek to a negative position. I fixed that and added some documentation and renamed some variables for clarity. Also tweaked the tests a little and added a new test case.


joelrebel (approved):
These changes seem fine to me

PR #142, opened 2024-05-07
Add format support to nvme WipeDisk

mmlb:
### What does this PR do

Adds support for wiping nvme disks using nvme format for those devices that do not support sanitize.


mmlb:
depends on #142

joelrebel:
> depends on #142

cyclic dependency alert!

turegano-equinix (changes requested):
we need to check if err is nil here

https://github.com/metal-toolbox/ironlib/pull/142/files#diff-fd841c72621ce6171ce69fa396aa58dff324510ec8b41655a9d029abbad2176eR247-R251

turegano-equinix (approved):

PR #158, opened 2024-05-24
More clean ups

mmlb:
#### Make the dmidecode utility act like the others

dmidecode was using a method in the 3rd party library to look up the binary.
I've changed it to act like the other utilities by supporting binary tool override via environment variable.

#### Get rid of OemComponents struct

This old setup isn't very generic, we'd have to grow the struct for each new manufacturer that has OEM devices.
It also doesn't seem very useful like it was.
We already know the manufacturer via Hardware.Vendor so no need to stutter by having `Dell` in the OEM struct.

#### Move oem specific things out of model/

These belong in utils since they are util specific values.
Now there's no need to export them, so I made them private.

#### Delete UpdateReleaseEnvironments

Its not used in the code base nor any I can find by GitHub search.

### Note

These are "breaking" changes, but ironlib is <1.0 so I think its fine.
I envision more breaking changes as I work on using ironlib in vogelkop more tightly and need a more ergonomic api.


utils/dmidecode.go
```diff
@@ -1,8 +1,10 @@
 package utils
 
 import (
+	"cmp"
```
joelrebel, line 4: TIL: cmp

joelrebel (approved):
Thanks for the cleanup and fixes

utils/dmidecode.go
mmlb: Yep cmp.Or is new in 1.22, looking forward to making more use of it!

PR #159, opened 2024-05-29
Add basic support for blkdiscard

ScottGarman:
blkdiscard is a pretty basic command, it simply gets run against a block device, and returns an exit code. We always include the --force flag.

### How has this been tested?

On an Equinix Metal server with several spare disks, I've modified the `examples/blkdiscard/main.go` to point to one of these disks and confirmed that:

* The command succeeds in the happy path, and wipes the disk device which I had previously created a filesystem on
* The command is correctly using the --force flag, and will discard blocks even on a mounted disk device
* The command fails when it tries to run against a non-existent disk device
* The command fails when I set the context timeout to too low of a value (e.g, 1 second) 

Add support for blkdiscard command

### DIFF
--- a/examples/blkdiscard/main.go
+++ b/examples/blkdiscard/main.go
@@ -1,0 +1,40 @@
+package main
+
+import (
+	"context"
+	"flag"
+	"time"
+
+	"github.com/metal-toolbox/ironlib/utils"
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	device  = flag.String("device", "/dev/sdZZZ", "disk to wipe using blkdiscard")
+	timeout = flag.String("timeout", (2 * time.Minute).String(), "time to wait for command to complete")
+)
+
+// This example invokes ironlib and runs blkdiscard on the disk /dev/sdZZZ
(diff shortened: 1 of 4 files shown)


mmlb (approved):

examples/blkdiscard/main.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/metal-toolbox/ironlib/utils"
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	device  = "/dev/sdZZZ"
+	timeout = 2 * time.Minute
+)
+
+// This example invokes ironlib and runs blkdiscard on the disk /dev/sdZZZ
```
mmlb, line 16: can you add an arg so this example is a little more useful? I did that for nvme-wipe and actually found some errors which was nice :D https://github.com/metal-toolbox/ironlib/blob/main/examples/nvme-wipe/main.go#L13

mmlb: thanks!

mmlb (approved):

PR #161, opened 2024-06-06
Some more clean ups 

mmlb:
#### Silences examples/inventory

I was getting corrupted stdout/json when I tried to run examples/inventory because flashrom would print to stdout and make it invalid json. I also got rid of inventory.json from the directory since it doesn't seem to belong in source control.

#### Refactor models files

Files had contents that did not match the filenames so I broke them out to more appropriately named files. Not a big deal if coming from an editor and using go-to-definition but annoying when trying to open a file from the cli and there was no obvious file.

#### Save LogicalName to Drive if we have it

No real reason to not store this, its actually very helpful when trying to act on Drives if it knows what its LogicalName/device-name is. There was a worry about alloy possibly uploading a value that isn't guaranteed to be stable from multiple boots but alloy only uploads specific fields and not the whole Drive, so we're ok.

#### Store discovered nvme capabilities in Drive.Capabilities

Similar to LogicalName, there's not really a good reason to drop this information and its necessary for clients to be able to choose appropriate actions so don't make them re-run the scan.

#### Clean up Hdparm capabilities a little bit

Just some small refactors for better readability. Moved some variable definitions closer to where they first get use so we don't have to keep the context in our minds unnecessarily. Also tweaked the way a common.Capability is instantiated so that they are indented and we avoid long blocks of similar looking texts.

N.A.

### The HW model number, product name this change applies to (if applicable)

N.A.

### The BMC firmware and/or BIOS versions that this change applies to (if applicable)

N.A.

N.A.

:eyes: 

* `Drive`s now carry the `LogicalName` if one is handy
* nvme `Drive`s now carry the discovered `Capabilities`


DoctorVin (approved):
Seems uncontroversial

PR #166, opened 2024-06-27
Add a loopback block device test for blkdiscard

ScottGarman:
### What does this PR do

This adds a unit test that uses a loopback device to verify that the blkdiscard code
does indeed wipe a block device.

This test depends on `losetup` `sha256sum` and `od`. All appear to be present and working in CI.

The new test is now run in CI.

N/A (just testing improvement)

### DIFF
--- a/utils/blkdiscard_test.go
+++ b/utils/blkdiscard_test.go
@@ -1,8 +1,10 @@
 package utils
 
 import (
+	"bytes"
 	"context"
 	"os"
+	"os/exec"
 	"testing"
 
 	"github.com/bmc-toolbox/common"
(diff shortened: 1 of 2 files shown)


utils/blkdiscard_test.go
```diff
@@ -27,3 +29,71 @@ func Test_blkdiscard(t *testing.T) {
...
+	f, err := os.Create(fileName)
+	assert.NoError(t, err)
+
+	// Fill the file with 1's
+	_, err = f.WriteAt(bytes.Repeat([]byte{1}, int(fileSize)), 0)
+	f.Close()
+	assert.NoError(t, err)
+	t.Cleanup(func() { os.Remove(fileName) })
+
+	t.Log("hi")
```
mmlb, line 59: still here

utils/blkdiscard_test.go
```diff
@@ -27,3 +29,71 @@ func Test_blkdiscard(t *testing.T) {
...
+	// Create a loopback device using losetup
+	losetupOutput, err := exec.Command("losetup", "--find").Output()
+	assert.NoError(t, err)
+	loopbackDevice := string(bytes.TrimSpace(losetupOutput))
+	t.Logf("DEBUG: loopbackDevice will be: %s", loopbackDevice)
+	_, err = exec.Command("losetup", "--find", fileName).Output()
```
mmlb, line 65: why run losetup --find twice? This is liable to toctou race and loopbackDevice being incorrect. Instead, you can run `losetup --associated filename` and it returns all the loopback devices attached to this file (hopefully only one :) )

mmlb (changes requested):

utils/blkdiscard_test.go
nshalman: Haha, whoops. Thanks!

utils/blkdiscard_test.go
nshalman: It was being weird on my machine. I'll try it your way.

utils/blkdiscard_test.go
nshalman: removed

utils/blkdiscard_test.go
nshalman: cleaned up.

mmlb (approved):
lgtm thanks!

PR #167, opened 2024-07-08
Add support for hdparm based drive wipe

mmlb:
### What does this PR do

Similar to what we're doing for NVME devices, this command supports
Sanitizing and Erasing and decides which function and parameters to use
depending on the capabilities reported by the drive.

Sanitize is preferred over erase. Cryptographic scramble is preferred of
block erase within Sanitize.

I reused the nvme sanitize/erase type since utils is all one big package
anyway to avoid confusion. If we ever split each utility into its own
package (as I want) then we'll just create a more specific type for
hdparm.

### DIFF
--- a/examples/diskwipe/main.go
+++ b/examples/diskwipe/main.go
@@ -66,18 +66,30 @@ func main() {
 	case "nvme":
 		wiper = utils.NewNvmeCmd(*verbose)
 	case "sata":
-		// Lets see if drive supports TRIM, if so we'll use blkdiscard
+		// Lets figure out the drive capabilities in an easier format
+		var sanitize bool
+		var esee bool
+		var trim bool
 		for _, cap := range drive.Capabilities {
-			if strings.HasPrefix(cap.Description, "Data Set Management TRIM supported") {
-				if cap.Enabled {
-					wiper = utils.NewBlkdiscardCmd(*verbose)
-				}
-				break
+			switch {
+			case cap.Description == "encryption supports enhanced erase":
+				esee = cap.Enabled
+			case cap.Description == "SANITIZE feature":
+				sanitize = cap.Enabled
+			case strings.HasPrefix(cap.Description, "Data Set Management TRIM supported"):
+				trim = cap.Enabled
 			}
 		}
 
-		// drive does not support TRIM so we fall back to filling it up with zero
-		if wiper == nil {
+		switch {
+		case sanitize || esee:
+			// Drive supports Sanitize or Enhanced Erase, so we use hdparm
+			wiper = utils.NewHdparmCmd(*verbose)
+		case trim:
+			// Drive supports TRIM, so we use blkdiscard
+			wiper = utils.NewBlkdiscardCmd(*verbose)
+		default:
+			// Drive does not support any preferred wipe method so we fall back to filling it up with zero
 			wiper = utils.NewFillZeroCmd(*verbose)
 
 			// If the user supplied a non-default timeout then we'll honor it, otherwise we just go with a huge timeout.
--- a/utils/hdparm.go
+++ b/utils/hdparm.go
@@ -2,13 +2,17 @@ package utils
 
 import (
 	"bufio"
+	"bytes"
 	"context"
+	"fmt"
 	"os"
 	"regexp"
 	"strings"
+	"time"
 
 	"github.com/bmc-toolbox/common"
 	"github.com/metal-toolbox/ironlib/model"
+	"github.com/sirupsen/logrus"
 )
 
 const (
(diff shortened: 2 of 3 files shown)


examples/diskwipe/main.go
```diff
@@ -66,18 +66,30 @@ func main() {
...
+		var esee bool
+		var trim bool
 		for _, cap := range drive.Capabilities {
-			if strings.HasPrefix(cap.Description, "Data Set Management TRIM supported") {
-				if cap.Enabled {
-					wiper = utils.NewBlkdiscardCmd(*verbose)
-				}
-				break
+			switch {
+			case cap.Description == "encryption supports enhanced erase":
```
joelrebel, line 75: It may be worth using `strings.EqualFold()` here

utils/hdparm.go
```diff
@@ -207,6 +211,136 @@ func (h *Hdparm) DriveCapabilities(ctx context.Context, logicalName string) ([]*
...
+	}
+
+	h.Executor.SetArgs("--yes-i-know-what-i-am-doing", "--sanitize-"+sanType, drive.LogicalName)
+	_, err = h.Executor.Exec(ctx)
+	if err != nil {
+		return err
+	}
+
+	// now we loop until --sanitize-status reports that sanitization is complete
+	for {
```
joelrebel, line 297: I suggest this loop be external to the library, so as a principle - the library always performs an action, returns the result and allows the caller to check the status of the action. 
This way we'll avoid situations where its unclear whats going on because the library hasn't returned for a while and a context cancellation doesn't cause it to return.

utils/hdparm.go
```diff
@@ -207,6 +211,136 @@ func (h *Hdparm) DriveCapabilities(ctx context.Context, logicalName string) ([]*
...
+	default:
+		return fmt.Errorf("%w: %v", errSanitizeInvalidAction, sanact)
+	}
+
+	verify, err := ApplyWatermarks(drive)
+	if err != nil {
+		return err
+	}
+
+	h.Executor.SetArgs("--yes-i-know-what-i-am-doing", "--sanitize-"+sanType, drive.LogicalName)
```
joelrebel, line 290: :'D

utils/hdparm.go
```diff
@@ -207,6 +211,136 @@ func (h *Hdparm) DriveCapabilities(ctx context.Context, logicalName string) ([]*
...
+		case strings.HasPrefix(cap.Description, "erase time:"):
+			eseu = strings.Contains(cap.Description, "enhanced")
+		}
+	}
+
+	l := logger.WithField("drive", drive.LogicalName)
+	if sanitize && cse {
+		// nolint:govet
+		l := l.WithField("method", "sanitize").WithField("action", "sanitize-crypto-scramble")
+		l.Info("wiping")
```
joelrebel, line 243: Can this log at debug level instead, it would be ideal to avoid info level logging in the library

utils/hdparm.go
```diff
@@ -207,6 +211,136 @@ func (h *Hdparm) DriveCapabilities(ctx context.Context, logicalName string) ([]*
...
+	var (
+		esee     bool
+		eseu     bool
+		sanitize bool
+		bee      bool
+		cse      bool
+	)
+	for _, cap := range drive.Capabilities {
+		switch {
+		case cap.Description == "encryption supports enhanced erase":
```
joelrebel, line 226: `strings.EqualFold()` would be ideal here as well.

utils/hdparm.go
mmlb: This is useful info that I do not want to hide behind both debug level and potentially lots of log messages. With both FillZeros usually taking a long time and both hdparm and nvme wipes introspecting the capabilities and potentially trying multiple wipe methods/actions I think logging these makes sense. If the user does not want to see them then they can pass in a logger that will filter out info messages. I guess the same could be said for doing at debug too :D. Ok you've convinced me :), but I'd like to do that in a follow up PR since the other WiperDrive functions are currently logging at info level and I'd like to do them all in one PR.

utils/hdparm.go
mmlb: same response:
this string is hard coded by hdparm in https://github.com/metal-toolbox/ironlib/blob/main/utils/hdparm.go#L192, so no need for equalfold

examples/diskwipe/main.go
mmlb: this string is hard coded by hdparm in https://github.com/metal-toolbox/ironlib/blob/main/utils/hdparm.go#L192, so no need for equalfold

utils/hdparm.go
mmlb: I added the loop because when I was testing this out manually `hdparm --sanitize-status` returns progress info (in a weird order) but when I've tried out the example it actually blocks until idle, so this loop might be able to go away. I'll go try and figure out why/how hdparm --sanitize-status does different behavior and may drop this loop. If we keep the loop in and timeout occurs then `.Exec` will return an error and we return from the function.

I'm trying to keep the WipeDrive interface simple to use and reason about. Callers ask us to wipe a drive, we wipe a drive or return an error. I also want to have it so that if we returned successfully then the caller is sure that the drive was wiped, we guarantee that by wrapping the underlying execs around `verify, err := ApplyWatermarks(drive)` and `verify()`.

So unless we want to rework how `WipeDrive` works its not really something to do in this PR. If we did want to do that it would also be a breaking change.

joelrebel: Yeah, I theres going to be cases like this and you can refactor iteratively.
In principle though, lets try to have the library return control to the caller asap and leave any polling to the caller.

utils/hdparm.go
joelrebel: Theres other code in ironlib thats logging at info level as well (that would be my fault as well).
Logging is not often the behaviour we want from a library, but I can see with destructive/long running actions like wiping drives this can be necessary. My suggestion was more of a guideline, feel free to change things around when it becomes noisy.

joelrebel (approved):

utils/hdparm.go
mmlb: Originally I was working on a workflow like `func DriveWipers() []DriveWiper` and we'd return each sub-method but scrapped it for now. I'd want some ways to get some metadata for each method like `func IsSecure() bool` and/or `func IsFast() bool` with the idea that the caller has more info and can make its own decisions, having it also need to poll for doneness and verify can make sense then too, but idk I still like the simple interface of WipeDrive either wipes the drive or returns an error, if the user wants concurrency they can do it at the disk level with goroutines.

utils/hdparm.go
mmlb: Agree about logging, I have the same feeling. FillZero's progress output was the push for logging in the wipe method and I just went with info since I was being explicit about passing in the logger instead of embedding/re-using previous one I figured it wasn't that bad of an idea, but yeah I'll follow up with a PR for changing this.

PR #169, opened 2024-07-09
Change Info log messages

mmlb:
### What does this PR do

Changes messages that are logged as `.Info` to either `.Debug` or `.Error` as appropriate. We should not be logging under `.Info` in a library.


joelrebel (approved):
👍

PR #179, opened 2024-08-30
Cleanup Dockerfile and convert non-distributable handling to ONBUILD

mmlb:
### What does this PR do

Bunch of minor (imo) clean ups to the Dockerfile and scripts/install-non-distributable.sh to start.
Then converts the whole fetching of non-distributable files from at ironlib image build time to `ONBUILD` and thus when consumers build their own images based off of ironlib's image.

This looks similar to how we used to do things before except there's better handling for skipping all of it by keeping what we do today of requiring a build-arg to opt in to proprietary deps. The old ONBBUILD setup used to check for presence of the fetcher script to decide to run the fetch. That was annoying because consumers would need a whole Dockerfile/repo setup just to have the script.

Build this docker image with/without the args to opt-in to proprietary binaries, there should be no difference between the 2 images. Lets assume you tag it `my-local-ironlib`.

Now do the same with/without build using the following Dockerfile:
```
FROM ghcr.io/metal-toolbox/vogelkop:v0.4.0 AS vogelkop

FROM my-local-ironlib
COPY --from=vogelkop /usr/sbin/vogelkop /usr/sbin/vogelkop
```

Running `docker build --build-arg INSTALL_NON_DISTRIBUTABLE=true` will fail and complain about missing AWS creds. Build again with the build-arg and it will complete. Also note that since no `ARG` is present in this example Dockerfile the secrets do not leak into the final image (verification left up to reader).


Dockerfile
```diff
@@ -1,94 +1,87 @@
...
-
-# copy ironlib wrapper binaries
-COPY --from=stage0 /usr/sbin/getbiosconfig /usr/sbin/getbiosconfig
-COPY --from=stage0 /usr/sbin/getinventory /usr/sbin/getinventory
+FROM almalinux:9-minimal AS base
+FROM base AS deps
 
 # import and install tools
 RUN curl -sO https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm
-RUN microdnf install -y --setopt=tsflags=nodocs crypto-policies-scripts && \
```
joelrebel, line 36: does the `tsflags=nodocs` not help as much here?

Dockerfile
```diff
@@ -1,94 +1,87 @@
...
+      microdnf install -y srvadmin-idracadm7; \
+    fi
 
 # update dependencies
-RUN microdnf update -y --setopt=tsflags=nodocs --setopt=install_weak_deps=0 \
-                       --setopt=keepcache=0 && microdnf clean all
+RUN microdnf update -y && microdnf clean all
 
 # install misc support packages
-RUN microdnf install -y --setopt=tsflags=nodocs --setopt=install_weak_deps=0 \
```
joelrebel, line 60: is there a reason we're dropping `--setopt=install_weak_deps=0` here, I'd like to be sure it doesn't affect the images as is

mmlb: So my original intent was to figure out a way to avoid cluttering up the microdnf calls (and also avoid an issue if forgotten). I meant to either export an env var or write into a config file and have that take care of clutter/memory. Before doing that I wanted to see what packages would be installed w/o the flags so I could verify the envvar/config file was being used. To my surprise same exact packages were added with/withouth the flags so I decided to just drop them.

Dockerfile
mmlb: not that I could tell, though I was looking for -docs packages to be installed like in other distros.

Dockerfile
mmlb: Hmm actually looks like I missed a dep that snuck in! Without install_weak_deps=0 ` libxcrypt-compat-4.4.18-3.el9.x86_64                                    appstream               90.0 kB` sneaks in! Let me try the config file approach then.

mmlb: Yep the config file approach works, I'll add weak_deps and docs too (just in case it changes in the future) to it:
```
oot@70866f3dc3d0 /]# microdnf install python 2>&1 | grep Installing
Installing:
 Installing:        7 packages
n
[root@70866f3dc3d0 /]# echo install_weak_deps=0 >>/etc/dnf/dnf.conf
[root@70866f3dc3d0 /]# microdnf install python 2>&1 | grep Installing
Installing:
 Installing:        6 packages
n
```

mmlb: updated, ptal

joelrebel (approved):
Thanks for fixing this up!

PR #194, opened 2025-01-21
Pin to DSU_24.11.11 to include srvadmin-idracadm7

turegano-equinix:
### What does this PR do
This PR updates the DSU repository URL to DSU_24.11.11 to ensure compatibility with the srvadmin-idracadm7 package, which is no longer available in the latest repository version. The DSU version is pinned as a workaround. A comment has also been added referencing a related Dell support case.


GraemeGregory:
This looks like the correct solution to me until we can implement an alternative to DSU!

ofaurax (approved):
Needed while Dell ponders what to do

PR #199, opened 2025-01-30
Force the formatting of a nvme namspace to avoid 10s delay

GraemeGregory:
### What does this PR do

Adds --force to the nvme format command

Without the --force the command waits for 10s for user input to cancel. As we are running this under automation there is no user to cancel and our intent is always to destroy the data.

go tests have been updated for the new argument.

https://equinixjira.atlassian.net/browse/BMS-885

n/a

### The HW model number, product name this change applies to (if applicable)

n/a

### The BMC firmware and/or BIOS versions that this change applies to (if applicable)

n/a

n/a

Format an nvme namespace, it will no longer prompt for 10s for human intervention.

Force the format of a nvme namespace to avoid 10s delay for human intervention.


turegano-equinix (approved):
LGTM

PR #203, opened 2025-02-03
Allow multiple wipes to run simultaneously

turegano-equinix:
### What does this PR do
It synchronizes the buffers every 60 seconds to release blocking syscalls from other ironib instances

use examples/diskwipe

Multiple disks wiped by ironlib

### DIFF
--- a/utils/fill_zero.go
+++ b/utils/fill_zero.go
@@ -9,6 +9,7 @@ import (
 
 	common "github.com/metal-toolbox/bmc-common"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
 )
 
 type FillZero struct {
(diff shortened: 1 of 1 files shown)


GraemeGregory (commented):
Isnt this PR a rather complex way of passing O_SYNC to the open() call?

turegano-equinix:
> Isnt this PR a rather complex way of passing O_SYNC to the open() call?

With O_SYNC, the speed drops dramatically.

GraemeGregory:
> > Isnt this PR a rather complex way of passing O_SYNC to the open() call?
> 
> With O_SYNC, the speed drops dramatically.

you are aware the data in either case only gets written during the sync operation. Have you done proper analysis that O_SYNC vs this have a realistic time different over the whole operation?

What size blocks are we writing, as if we are smaller than the natural size of the storage then more time is lost re-writing blocks.

turegano-equinix:
The problem is with the discovery code (it tries to get information from other disks), we'll open a new branch to fix that

PR #204, opened 2025-03-10
lshw: build serial number for NICs if needed

ofaurax:
### What does this PR do

Build a serial number for NICs without one.

Reports NICs without serial number. Before, they were ignored.


ofaurax:
We should get the MAC as serial

